PR #13, opened 2016-02-24
Add style option setext_with_atx_closed to header style rule MD003

ciwchris:
Atx-style headers which are closed can now be used with setext style
headers, when at header level 3 or more.


lib/rules.js
```diff
@@ -197,9 +197,11 @@ module.exports = [
           style = styleForToken;
         }
         if ((styleForToken !== style) &&
-            !((style === "setext_with_atx") &&
+            !((style === "setext_with_atx" || style === "setext_with_atx_closed") &&
               (/h[12]/.test(token.tag) && (styleForToken === "setext")) ||
-               (/h[^12]/.test(token.tag) && (styleForToken === "atx")))) {
+              (/h[^12]/.test(token.tag) &&
+                  (styleForToken === "atx" || styleForToken === "atx_closed")))
```
ciwchris: This if statement is getting a bit hairy, but it looked like you kept all of the rule logic together, so I didn't want to break it out into its own function.

DavidAnson:
Awesome, thank you!! I'll review the PR in detail tomorrow. At first glance, everything looks right. :)

lib/rules.js
DavidAnson: Cool, this is what I would have done.

DavidAnson: Nitpick: It looks like line 202 lost a space of indentation (as rendered in the PR on GitHub)?
My convention is usually to go in one space for each level of nesting - so 203 would go 2 spaces to the left and 202 1 to the right if I'm reading this correctly.

doc/Rules.md
```diff
@@ -77,8 +78,8 @@ Be consistent with the style of header used in a document:
 
     ## ATX style H2
 
-The setext_with_atx doc style allows atx-style headers of level 3 or more in
-documents with setext style headers:
+The setext_with_atx and settext_with_atx_closed doc styles allows atx-style
```
DavidAnson: "allows" -> "allow"

lib/rules.js
ciwchris: Line 201 and 202, and actually 200 as well, are all siblings, so I lined them all up on their opening paren. Is it still expected that line 202 should be indented one space to the right?

Regarding 203, sorry, my default tab is four spaces. Line 203 is a nested condition. Would you like it 1 or or 2 spaces in from the previous line?

DavidAnson: As I look at this more, I'm not even sure I followed my own guidelines with the original code. :) For line 203, I'd prefer it one space in from the previous line. And please don't spend any more time on indentation - I feel bad about the distraction!

DavidAnson:
FYI, I added a test which caught an edge case in the overly-complicated if statement you called out. So I refactored to make it more clear and fix the problem. Thanks again for the contribution!

PS - Your indentation was perfect. :)

PR #16, opened 2016-03-13
Adding link to remark-lint

igorshubovych:
Adding link to related project Remark-Lint:
https://github.com/wooorm/remark-lint


DavidAnson:
remark-lint looks like a nicely done project, but it's not really *related* to markdownlint in the same way that the existing "Related" section items are (i.e., they use markdownlint directly or support the same set of rules in another language).

Perhaps a section "Other markdown linters" could be added to the Readme - but I wouldn't want to have to spend time maintaining that kind of list over time. Is there something particular about remark-lint that you feel is relevant to this audience?

To be clear, I don't have a strong feeling here - but I'd like to understand the change better before pulling it in. Thanks!

igorshubovych:
I agree with your thoughts.
It is different. As I understood `mdl` was the inspiration for the `markdownlint`.

PR #54, opened 2017-04-06
Split rule documentation into separate files

pascalberger:
Split rule documentation into separate files. This allows to easily link to individual rules, compared to before where there was not only the rule number, but also a description in the anchor. This allows also to automatically generate links to rule documentation if rule number is known.


DavidAnson:
Thank you!

It seems to me this change has some advantages as well as some disadvantages.

Advantages:

- I like that you added links from the README to each of the rules!
- I like the idea of making rules easier to link to from other pages and tools by simplifying the external URL!

Disadvantages:

- When creating the Node.js implementation of markdownlint from the original Ruby project, one of my goals was to preserve as much of the file/test structure as I could so it would be easy to transfer changes back and forth. Breaking up the RULES file as you've done seems like it would make future merges more time-consuming.
- The Travis continuous integration tests appear to be broken with this change.
- It would be best to update the "docs" tests to verify the new README links.

I had a couple of ideas that might keep the spirit of this change without some of the downsides:

- Update the README to link to the each of the rules by their current RULES file links.
- Find a way to create shorter links to each rule without splitting the current RULES file.

DavidAnson:
My commit above implements the desired behavior without splitting Rules.md into multiple files (see my comment above for why).

PR #69, opened 2017-07-27
Add support to run markdownlint with pre-commit

Dunedan:
This PR adds the ability to use `markdownlint` as [pre-commit](http://pre-commit.com/) hook.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,6 +1,6 @@
 -   id: markdownlint
     name: markdownlint
-    description: "Check the style of Markdown/Commonmark files"
+    description: "Checks the style of Markdown/Commonmark files."
     entry: markdownlint
     language: node
     types: [markdown]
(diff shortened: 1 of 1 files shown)


DavidAnson:
Thank you, this looks neat! I haven't used `pre-commit` before, so I have a few questions:

1. My reading of the `entry` field for Node packages is that it points to an executable or script to run, but `markdownlint` doesn't include one (that's what `markdownlint-cli` is for). So how does this work?
2. Why is it necessary to specify 0.15.0 as the minimum version? Is this required?
3. Very minor, and please ignore if it goes against standards, but would you mind changing to "Checks" in the `description` and adding a period at the end?

Thank you again!

Dunedan:
1. A right. Turns out I had [markdown-cli](https://github.com/igorshubovych/markdownlint-cli) installed globally, that's why it worked. I guess I should open this PR then against `markdown-cli` instead.
2. 0.15.0 introduced the `types` attribute. Before you had to select the files to check via regexes matching the file name.
3. Already adjusted :smile:

DavidAnson:
Okay, thanks, it sounds like this needs to be done for the `markdown-cli` project instead. I've considered bundling a CLI here, but like keeping it separate so far.

Dunedan:
All right. I'll open a PR against `markdownlint-cli` then. Sorry for bothering you.

PR #86, opened 2017-10-30
Moved rules into their own files

### LINKED ISSUE #83: Split rules into their own files?
https://github.com/DavidAnson/markdownlint/blob/master/lib/rules.js is huge. It would be easier to develop if the rules were in their own files.

JoshuaKGoldberg:
Everything under `lib/rules/*` is a rules file (with the name of the rule in camelCase), re-exported into an array in `lib/rules.js`.
Moved the regular expressions from `lib/rules.js` to `lib/expressions.js`, and the rest of the variables into `lib/shared.js`.

Fixes #83.


JoshuaKGoldberg:
> Coverage decreased (-79.2%) to 20.8% 

It's 100% locally :(

Issue #75, opened 2017-10-10
Add rule to enforce image alt text.


sstedman:
Add a rule to detect and throw error for empty image alt text:

```
![](/path/to/image)
```

Add corresponding tags for `image` (and `accessibility`?).

sstedman:
Super, thanks! 

Regarding the last sentence: if it makes sense, I thought this rule could also use some new tags that don't yet exist in the  https://github.com/DavidAnson/markdownlint#tags list.

DavidAnson:
Understood, good idea!

KvanTTT:
I'm working on [MarkConv](https://github.com/KvanTTT/MarkConv) utility that can convert markdowns of different kinds (GitHub, VisualCode, Habrahabr). Also, it warns about some style issues. Maybe it will be interesting and helpful for you.

PR #101, opened 2017-12-30
Adding a new rule, no-alt-text as per Issue #75

DuncanmaMSFT:
Request was for a new rule to check for missing alt text in markdown image syntax


DavidAnson:
This looks great, thanks! I’ll have time to go over it in more detail soon - here are a couple of things I noticed during a first pass:

+ `no_alt_text.json` should be unnecessary because rules default to being enabled
+ Could you please sort the new tag names in README and RULES?
+ Could you please add a test case for the [reference style](https://daringfireball.net/projects/markdown/syntax#img) of specifying images?
+ If you run the script to generate the schema file, it will sort the new entries consistently
+ Could you please add “(fixes #75)“ to the end of the commit message so that issue is automatically closed?
+ Very picky, but could you please use hyphens instead of underscores for the name `no_alt_text.md`?

Thanks again, I appreciate the help!

DuncanmaMSFT:
Thanks for the feedback, I'll make those changes as soon as I can

DavidAnson:
Probably final request: Please squash all commits so the PR is all in a single commit with the relevant message. Thanks!

DavidAnson:
Can you tell me what scenario this part of the code handles?

`if (token.content === "") {`

Code coverage did not decrease, so I think the `else` is getting run - I’m just not sure why...

DavidAnson:
Merged into branch `next` via https://github.com/DavidAnson/markdownlint/commit/29e64d86dd14a3b4f341ce9648f0f675aa486591.

DuncanmaMSFT:
I know this is already closed, but I wanted to answer your question. It seems that the markdown processor, in the reference style case, puts the alt text into the content portion of the token. In the inline image syntax, it ends up in an attribute, and the content value is empty. So, to cover both forms of syntax I consider either a non-empty content value or an 'alt' attribute value to be acceptable.

DavidAnson:
Thanks! I did some experimentation and added more test cases, but it seemed to me that the only thing that was necessary was the check for content. In my follow up commit, I removed the alt code entirely. If you know of a scenario that requires that code, please do let me know  because I may have broken it. I believe there is test coverage for both of the flavors you mention above, so perhaps there is a specific variant of one of them that needs the extra check?

PR #111, opened 2018-03-17
Add a link to linter-node-markdownlint

guillaumearm:
There is a plugin that allow to use markdownlint with the [atom linter](https://atom.io/packages/linter)

For me, I just ran `apm install linter-node-markdownlint` and restart atom.
Everything works fine.


DavidAnson:
Great find, thank you!

PR #115, opened 2018-03-19
Change "header" to "heading" across the library

### LINKED ISSUE #113: Header vs. heading
This library seems to misuse the word "header" a lot, when in fact, as far as I can see, it always actually refers to a "heading". This is a common mistake all over the web (and a personal pet peeve of mine), but both the [HTML standard](https://www.w3.org/TR/2011/WD-html-markup-20110525/elements.html) and the [CommonMark spec](http://spec.commonmark.org/0.28/#atx-headings) correctly refer to these tags as headings, whereas a header is something completely different, and even has its own [HTML tag](https://www.w3.org/TR/2011/WD-html-markup-20110525/header.html#header).

I realize this is because the original https://github.com/markdownlint/markdownlint library does the same, and I've created an issue there as well - markdownlint/markdownlint#194 - but that seems to be a less maintained repo, as it hasn't seen new commits in over half a year - this package seems to be more popular, especially as Node-based build systems are more common these days.

Apart from the fact that it's technically incorrect, I've also seen it cause actual confusion in practice. I'm pretty sure this change would be a BC-break, because it would require changing the naming of tags and aliases for the affected rules, but are you open to fixing the naming inconsistency at some point in the future? Furthermore, are you okay with a moving the direction of this project away from the original library by implementing a change like this?

levacic:
This should be backward compatible, as all "header" aliases are still
available, though documented as discouraged for future use.

Fixes #113


levacic:
@DavidAnson As you wanted backward compatibility, simply bulk-replacing stuff wasn't an option, so it took a bit of time to get everything right - especially as some tests were a bit weird. Specifically the one that takes the rule function as a string and extracts `params.config.*` to see which parameters were referenced - this forces MD013 to have code like this:

```js
    var headings = params.config.headings;
    if (headings === undefined) {
      headings = params.config.headers;
    }
```

whereas something like this might be more readable (although it breaks line length):

```js
var headings = (params.config.headings !== undefined) ? params.config.headings : params.config.headers;
```

Either way, I realize the PR is huge, and I'm not 100% happy with the docs - but let me know what you think and if you'd like some additional changes. It's not trivial to document everything nicely due to how tightly coupled the docs are to the implementation via the tests.

DavidAnson:
Thanks, I’ll have a look after the current round of changes is complete!

Maybe: var h = header || heading;

levacic:
@DavidAnson That was my first instinct as well, but it doesn't work if someone passes `false` for `header` - the only correct way here is to check for `undefined`.

The changes _are_ complete, this is (supposed to be) a fully-functioning PR, everything has been updated including the tests, and everything is passing (I've executed the relevant `scripts` from `package.json`: `test`, `test-cover`, `test-extra`, `lint` and `build-config-schema`).

Do you have any specific changes you'd like me to make?

DavidAnson:
Sorry, I didn’t mean *your* changes, I meant *mine*. :) I have fixes already under way to this project, then two dependent project updates to do before I come back to this.

Regarding the snippet, I’m not sure it’s necessary to support overlapping/conflicting configuration, but it’s minor either way.

levacic:
Ah, that makes sense.

As for the overlapping configuration, I figured it was the only way to provide reliable BC - otherwise it would get very confusing (being able to use old rule names but forced to update the rule configuration).

If you'd rather not have that, I suggest just dropping `header` completely and releasing a new major version - it'll be easier to maintain in the long run anyway.

DavidAnson:
I think you are right to support old and new name for configuration, thanks. Your example of false/undefined sounded like you were trying to support someone who used old and new names in the *same* config - and that seemed unnecessary. But thinking about it more, I see that you are right because false falling back to undefined changes the meaning. So thanks again. :)

levacic:
No worries - ideally, we'd disallow somehow using both at the same time (or throw deprecation warnings for using `header` in any context, if you do wish to deprecate this at some point), but I wasn't sure how to go about doing this - if you have any ideas I'd be glad to help.

Also let me know if you want me to rebase the changes or something (not sure what your preferred Git workflow is for this) once you're done with yours - though there doesn't seem to be any conflicts yet so that's a good sign :)

README.md
```diff
@@ -46,9 +46,9 @@ playground for learning and exploring.
 
 ## Rules / Aliases
 
-* **[MD001](doc/Rules.md#md001)** *header-increment* - Header levels should only increment by one level at a time
-* **[MD002](doc/Rules.md#md002)** *first-header-h1* - First header should be a top level header
-* **[MD003](doc/Rules.md#md003)** *header-style* - Header style
+* **[MD001](doc/Rules.md#md001)** *heading-increment, header-increment* - Heading levels should only increment by one level at a time
```
DavidAnson: There is commented-out code in `markdownlint.js` to generate this. I don’t see that changed here, but it would be good to do for future updates.

README.md
```diff
@@ -77,19 +77,23 @@ playground for learning and exploring.
...
-* **[MD043](doc/Rules.md#md043)** *required-headers* - Required header structure
+* **[MD043](doc/Rules.md#md043)** *required-headings, required-headers* - Required heading structure
 * **[MD044](doc/Rules.md#md044)** *proper-names* - Proper names should have the correct capitalization
 * **[MD045](doc/Rules.md#md045)** *no-alt-text* - Images should have alternate text (alt text)
 
 See [Rules.md](doc/Rules.md) for more details.
 
+> All rules with `heading` as part of their name are also available as `header`
+> aliases, e.g. `heading-increment` is also available as `header-increment`.
+> This is provided for backward compatibility, though its usage is discouraged.
```
DavidAnson: The use of `header` is deprecated and provided for backward-compatibility only.

README.md
```diff
@@ -101,6 +105,8 @@ See [Rules.md](doc/Rules.md) for more details.
 * **code** - MD014, MD031, MD038, MD040
 * **emphasis** - MD036, MD037
 * **hard_tab** - MD010
+* **headings** - MD001, MD002, MD003, MD018, MD019, MD020, MD021, MD022, MD023,
```
DavidAnson: The code to generate this will sort automatically.

doc/Rules.md
```diff
@@ -416,7 +416,9 @@ Tags: line_length
 
 Aliases: line-length
 
-Parameters: line_length, code_blocks, tables, headers (number; default 80, boolean; default true)
+Parameters: line_length, code_blocks, tables, headings, headers (number; default 80, boolean; default true)
+
+> If a `headings` parameter is not provided, `headers` will be checked instead for backward compatibility.
```
DavidAnson: ... provided, the deprecated `headers` parameter will be used instead.

doc/Rules.md
```diff
@@ -1343,16 +1345,18 @@ But non-empty fragments will not:
...
+Tags: headings, headers
 
-Tags: headers
+Aliases: required-headings, required-headers
 
-Aliases: required-headers
+Parameters: headings, headers (array of string; default `null` for disabled)
 
-Parameters: headers (array of string; default `null` for disabled)
+> If a `headings` parameter is not provided, `headers` will be checked instead for backward compatibility.
```
DavidAnson: Same comment as above

lib/md043.js
```diff
@@ -5,12 +5,12 @@
...
-  "names": [ "MD043", "required-headers" ],
-  "description": "Required header structure",
-  "tags": [ "headers" ],
+  "names": [ "MD043", "required-headings", "required-headers" ],
+  "description": "Required heading structure",
+  "tags": [ "headings", "headers" ],
   "function": function MD043(params, onError) {
-    var requiredHeaders = params.config.headers;
-    if (requiredHeaders) {
+    var requiredHeadings = params.config.headings || params.config.headers;
```
DavidAnson: Does this need the longer form as well?

schema/build-config-schema.js
```diff
@@ -139,8 +139,13 @@ rules.forEach(function forRule(rule) {
           "type": "boolean",
           "default": true
         },
+        "headings": {
```
DavidAnson: Same comment about being sorted when generated. See npm scripts for command to do so.

DavidAnson: Oh, oops, this is not the generated file. Then please sort manually. :)

test/code_block_dollar_fence-empty.md
```diff
@@ -10,5 +10,5 @@
 ```
 
 ```spaces
-
```
DavidAnson: Another diff removed unnecessary trailing spaces, but I think these spaces should not be removed.

test/long-heading-exceptions.md
```diff
@@ -1,20 +1,20 @@
-# Header
+# Heading
 
 Text
 
-## Header header header header header header header header header header header header header header header header
+## Heading heading heading heading heading heading heading heading heading heading heading heading heading heading heading heading
```
DavidAnson: LOL

README.md
```diff
@@ -46,9 +46,9 @@ playground for learning and exploring.
 
 ## Rules / Aliases
 
-* **[MD001](doc/Rules.md#md001)** *header-increment* - Header levels should only increment by one level at a time
-* **[MD002](doc/Rules.md#md002)** *first-header-h1* - First header should be a top level header
-* **[MD003](doc/Rules.md#md003)** *header-style* - Header style
+* **[MD001](doc/Rules.md#md001)** *heading-increment, header-increment* - Heading levels should only increment by one level at a time
+* **[MD002](doc/Rules.md#md002)** *first-heading-h1, first-header-h1* - First heading should be a top level heading
```
DavidAnson: Convention in result output is aaa/bbb, so I think that would be good to use here also.

DavidAnson: (Instead of ‘,’)

README.md
DavidAnson: To be clear, I suggest replacing the last sentence.

doc/Rules.md
DavidAnson: Again, I am rewording the last part.

levacic:
I'll be able to make the changes sometime over the weekend - if you want to merge this sooner, feel free to apply them yourself.

Thanks for going over the whole thing and providing feedback, PRs like this can be tricky to review confidently!

Cheers

DavidAnson:
Merged to `next` branch via f4f4728e2cb147dcf107bb683e714dc52e8754b3.

levacic:
Eh, just got notifications about this being updated and merged by yourself - sorry for not getting to it, life/work got into way, as you might imagine :)

Thanks for completing it and merging!

DavidAnson:
No worries, appreciate the careful changes!

PR #123, opened 2018-05-24
docs: Point to markdownlint-rules keyword on NPM

nschonni:
There doesn't look to be any results right now, but might be good for the future.
Not sure if the link should cover less of the text


DavidAnson:
😄 I had the same idea, but wanted to wait for at least one result before doing this. Maybe your link checker?

DavidAnson:
Let’s do this now. If you could please rebase onto the `next` branch, I will accept the PR.

DavidAnson:
Merged to `next` via 5a73fbb10009cfefde76c3275d8901e97105709d.

nschonni:
Thanks, sorry I didn't swap branches. Looks like just doing the `Edit` from the top was all that was needed since there where now other commits

Issue #127, opened 2018-06-22
Documentation on writing tests?


JoshuaKGoldberg:
It's not clear from the files under `test/` in this repo what the steps are to CRUD tests. Could you write some documentation on this please?

DavidAnson:
Tests are run by `npm test` in the usual manner. All tests pass on Windows (my environment, CRLF) and Linux (Coveralls, LF, https://travis-ci.org/DavidAnson/markdownlint/builds/384576202). The test runner is the popular `nodeunit` (https://github.com/caolan/nodeunit) and test cases are found in `test/markdownlint-test.js`. Additionally, `test/*.md` files are enumerated, linted, and pass/fail according to whether every line that triggers a warning has corresponding "{MD###}" text. For example, the line `#### Heading 2 {MD001}` is expected to trigger `MD001`. For cases where the marker can not be on the same line, the syntax "{MD###:#}" can be used to identify a different line number. If the `sometest.md` file needs custom configuration, a `sometest.json` file is used to provide a custom `options.config` for that test. For code coverage, run `npm run test-cover` and for linting, run `npm run lint`.

PR #128, opened 2018-06-22
Added md046 sentences-per-line

JoshuaKGoldberg:
Complains if a line contains a second period followed by a space and an [A-Z|a-z] character. Ignores <code>```</code> and <code>`</code> code blocks.

My build locally fails with `76` commits on `master`, so I couldn't easily figure out how to add tests. Filed #127.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,19 +16,20 @@ npm install markdownlint --save-dev
 ## Overview
 
 The [Markdown](https://en.wikipedia.org/wiki/Markdown) markup language is
-designed to be easy to read, write, and understand. It succeeds - and its
-flexibility is both a benefit and a drawback. Many styles are possible, so
-formatting can be inconsistent. Some constructs don't work well in all
-parsers and should be avoided. The [CommonMark](http://commonmark.org/)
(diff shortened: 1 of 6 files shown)


DavidAnson:
This looks like it might be a contentious rule, reporting failures in most scenarios (think of it like a significant breaking change). I agree with the sentiment and often use the "one sentence per line" practice myself, but experience has taught me to be careful about making the linter too opinionated.

What I'd prefer is to offer this as a custom rule. That allows anyone who wants to opt in without imposing it on everyone. Here's the documentation for custom rules, they're the same as internal rules (though they don't have access to all the helper methods): https://github.com/DavidAnson/markdownlint/blob/master/doc/CustomRules.md

JoshuaKGoldberg:
That sounds good to me, thanks for the documentation! Once I finish that work I'll likely ask about making a list somewhere in this repo of community rules.

JoshuaKGoldberg:
Following up: https://github.com/JoshuaKGoldberg/sentences-per-line exists. 😄

PR #154, opened 2018-11-14
feat: Add pre-commit hook

erdtsksn:
"As long as your git repo is **an installable package (gem, npm, pypi, etc.)** or exposes an executable, it can be used with pre-commit" This line comes from "See also" link. So we can use this repository instead of "markdownlint-cli".

### Why I prefer this repository instead of markdown-cli?
Because it is easier to add arguments to pre-commit hooks.

Example hook:
```
repos:
  - repo: https://github.com/erdtsksn/markdownlint
    rev: master
    hooks:
      - id: markdownlint
        args: ["--rules=~MD013"]
```
---
Related: https://github.com/DavidAnson/markdownlint/pull/69
See also: https://pre-commit.com/#new-hooks

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,7 @@
+- id: markdownlint
+  name: Markdownlint
+  description: "Checks the style of Markdown/Commonmark files."
+  entry: markdownlint
+  language: node
+  types: [markdown]
+  minimum_pre_commit_version: 0.15.0


DavidAnson:
Thanks! But I actually have the same question as last time this cane up. Could you please have a look at https://github.com/DavidAnson/markdownlint/pull/69 and let me know if this is the same situation?

erdtsksn:
This is the same feature. 

About your concern;
- It is not required to have an executable.

I haven’t install markdownlint globally. 
I haven’t install markdownlint-cli globally.
I have installed nodejs and python globally. 

I have test this pre-commit with my fork on my machine and it works.

If you have a corncern, I can test it using a clean machine.

DavidAnson:
It sounds like you set this up right and checked the right things and I trust you, but your example passes `--rules` as args and I don’t know what code in _this_ project would respond to that?

erdtsksn:
I think I have messed up a little. Let me check this pull request with a clean installation. It will take a little time because I need to reinstall my OS.

I have found the project that have these arguments. 

See: https://github.com/markdownlint/markdownlint/blob/master/docs/configuration.md#specifying-which-rules-mdl-processes

erdtsksn:
I have tested it with a clean installation. It **doesn't** work. Sorry about that.

For not I have to use [ruby version](https://github.com/markdownlint/markdownlint) because it supports arguments.

DavidAnson:
Thanks for looking into this!

PR #212, opened 2019-08-06
Rules: update MD032 removing deprecated paragraph

### LINKED ISSUE #211: Rule MD032 is not correctly checked
Rule MD032 does not seem to be correctly enforced when text is present on a line immediately *following* a list. See the example below.

File `test.md`:

    # Title

    * This is
    not okay

File `test.js`:

    const markdownlint = require("markdownlint");

    markdownlint({"files": ["test.md"]}, (_, res) => {
        console.log(res);
    });

Output:

    Results { 'test.md': [] }

Expected output:

    // Something like this:
    Results {
        'test.md': [
            {
                lineNumber: 4,
                ruleNames: [Array],
                ruleDescription: 'Lists should be surrounded by blank lines',
                // ...
            }
        ]
    }
    

---

Node version: 12.7.0
npm version: 6.10.2
`markdownlint` version: 0.16.0

mebeim:
Closes #211.

Also, I was wondering about this paragraph:

> This rule is triggered when lists (of any kind) are either not preceded **or not
followed by a blank line**
>
> ```markdown
> Some text
> * Some
> * List
> 
> 1. Some
> 2. List
> Some text
> ```

Doesn't the fact that "lazy continuation" is allowed basically contradict the second part of this sentence? In other words, would a final list item *not* followed by a blank line always be treated as lazy continuation and therefore never trigger this rule?

### DIFF
--- a/doc/Rules.md
+++ b/doc/Rules.md
@@ -1095,17 +1095,6 @@ Some text
 Rationale: Aside from aesthetic reasons, some parsers, including kramdown, will
 not parse lists that don't have blank lines before and after them.
 
-Note: List items without hanging indents are a violation of this rule; list
-items with hanging indents are okay:
-
-```markdown
-* This is
-not okay
-
-* This is
-  okay
-```
-
 <a name="md033"></a>
 
 ## MD033 - Inline HTML


DavidAnson:
I’d say the text is technically correct because it refers to lists and a lazy continuation extends the list. But I agree lazy confuses the issue here. Maybe this would be a better example?

    1. Some
    2. List
    # Heading

DavidAnson:
Could you please add “(fixes #211)” to the end of the first line of the commit message?

mebeim:
@DavidAnson sure thing, done.

DavidAnson:
Thanks!

Issue #242, opened 2019-12-29
MD007 ul-indent doesn't work when MD006 ul-start-left is disabled


jthomerson:
I like to have all my bullet lists always start indented with the standard level of indentation (in my case, three spaces). So, I had to disable `ul-start-left`, but then when I set the indentation of my lists (MD007), I get an off-by-one [nesting level] error for every single item in every list.

Consider this example:

```
jrthomerson@mine ~/code/markdownlint-example
[ master ] $ cat test.md
# Test

   * item 1
   * item 2
      * item 2.1
      * item 2.2
         * item 2.2.1
      * item 2.3
   * item 3
jrthomerson@mine ~/code/markdownlint-example
[ master ] $ cat .markdownlint.json
{
   "default": true,
   "line-length": {
      "line_length": 90
   },
   "no-multiple-blanks": { "maximum": 2 },
   "hr-style": { "style": "---" },
   "code-block-style": { "style": "fenced" },
   "code-fence-style": { "style": "backtick" },
   "ul-start-left": false,
   "ul-indent": { "indent": 3 },
   "no-alt-text": false
}
jrthomerson@mine ~/code/markdownlint-example
[ master ] $ ./node_modules/.bin/markdownlint ./test.md
./test.md:3 MD007/ul-indent Unordered list indentation [Expected: 0; Actual: 3]
./test.md:4 MD007/ul-indent Unordered list indentation [Expected: 0; Actual: 3]
./test.md:5 MD007/ul-indent Unordered list indentation [Expected: 3; Actual: 6]
./test.md:6 MD007/ul-indent Unordered list indentation [Expected: 3; Actual: 6]
./test.md:7 MD007/ul-indent Unordered list indentation [Expected: 6; Actual: 9]
./test.md:8 MD007/ul-indent Unordered list indentation [Expected: 3; Actual: 6]
./test.md:9 MD007/ul-indent Unordered list indentation [Expected: 0; Actual: 3]
```

PR #243, opened 2019-12-29
fix: allow for first level indentation of lists (#242)

jthomerson:
This is a proposed fix for #242. It should maintain backwards-compatibility, while allowing users like me to achieve our goal as well. If there's a different way you'd like it implemented, I'm happy to rework it with your direction.

### DIFF
--- a/lib/md007.js
+++ b/lib/md007.js
@@ -11,12 +11,14 @@ module.exports = {
   "description": "Unordered list indentation",
   "tags": [ "bullet", "ul", "indentation" ],
   "function": function MD007(params, onError) {
-    const optionsIndent = params.config.indent || 2;
+    const indent = params.config.indent || 2;
+    const startIndented = !!params.config.start_indented;
     flattenedLists().forEach((list) => {
       if (list.unordered && list.parentsUnordered) {
         list.items.forEach((item) => {
           const { lineNumber, line } = item;
-          const expectedIndent = list.nesting * optionsIndent;
+          const expectedNesting = list.nesting + (startIndented ? 1 : 0);
+          const expectedIndent = expectedNesting * indent;
           const actualIndent = indentFor(item);
           let range = null;
           let editColumn = 1;
--- a/schema/markdownlint-config-schema.json
+++ b/schema/markdownlint-config-schema.json
@@ -230,6 +230,11 @@
           "description": "Spaces for indent",
           "type": "integer",
           "default": 2
+        },
+        "start_indented": {
+          "description": "Whether to indent the first level of the list",
+          "type": "boolean",
+          "default": false
         }
       },
       "additionalProperties": false
@@ -246,6 +251,11 @@
           "description": "Spaces for indent",
           "type": "integer",
           "default": 2
+        },
+        "start_indented": {
+          "description": "Whether to indent the first level of the list",
+          "type": "boolean",
+          "default": false
         }
       },
       "additionalProperties": false
--- a/doc/Rules.md
+++ b/doc/Rules.md
@@ -273,7 +273,7 @@ Tags: bullet, ul, indentation
 
 Aliases: ul-indent
 
-Parameters: indent (number; default 2)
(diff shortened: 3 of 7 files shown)


DavidAnson:
Thank you! I started to review the PR, but the fact that the new parameter is a Boolean got me thinking: could this change be made without a parameter at all? As you note in the issue, MD006 already works for people who want to force a left-most start, so since your implementation assumes an indented start will be the same number of spaces as the indent, maybe we can just allow that as well? To work this way, I think the rule would need to track whether the list started out indented or not, but that doesn’t seem too different than what you’re doing already and doesn’t seem like it would complicate the logic much further. What do you think?

jthomerson:
Hmmm ... when I first read it, I liked your idea. But then I started thinking about how I'd implement it, and I realized that there's something I'd like to enforce that I don't think would be enforced by your proposal: I _always_ want the lists to be indented one level. With your suggestion, if I understood it correctly, someone could start their list not indented (at char zero of each line), and that would pass because we'd have to allow _either_ 0 or one level.

Is there a way to make MD006 enforce that lists _do_ start indented? Or should that be enforced here in MD007? Or a separate rule altogether?

DavidAnson:
Let’s do it here. I’ll review the PR now...

test/list-indentation-first-line-indented.md
```diff
@@ -0,0 +1,9 @@
+# Test
```
DavidAnson: I’ve started using the file name as the top-level heading.

test/list-indentation-first-line-indented.md
```diff
@@ -0,0 +1,9 @@
...
+
+   * item 1
+   * item 2
+      * item 2.1
+      * item 2.2
+         * item 2.2.1
+      * item 2.3
+   * item 3
```
DavidAnson: Please also add an example list without indent here to verify it’s reported as a violation. And make sure some other file tests for indent being used without the parameter as a violation - or add one.

schema/markdownlint-config-schema.json
```diff
@@ -230,6 +230,11 @@
           "description": "Spaces for indent",
           "type": "integer",
           "default": 2
+        },
+        "first_level_indented": {
+          "description": "Whether the first level should be indented, or start at zero",
```
DavidAnson: These are usually pretty terse. How about “Whether to indent the list”?

lib/md007.js
```diff
@@ -12,11 +12,13 @@ module.exports = {
   "tags": [ "bullet", "ul", "indentation" ],
   "function": function MD007(params, onError) {
     const optionsIndent = params.config.indent || 2;
+    const optionsLevel1Indent = (params.config.first_level_indented === true);
```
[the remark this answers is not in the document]
jthomerson: So, I renamed the var (I'd had that different name just because of the length, which we solved by shortening the config param name).

But, I didn't make it use loose equality. Since there's a schema that defines the config as undefined or boolean, I figured it was safe / better to leave it as strict equality. But, let me know if you prefer the loose equality and type coercion that goes along with it.

jthomerson:
@DavidAnson I believe I address all your comments. Let me know if I missed anything or you have further requests. Thanks!

lib/md007.js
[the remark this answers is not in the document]
DavidAnson: Here is the pattern I prefer to use: https://github.com/DavidAnson/markdownlint/blob/5634b317d6e152e2708c0cc1ecd12467101a9edb/lib/md013.js#L32

I see now that I wasn’t entirely consistent here because the existing variable has an “options” prefix. If you could please update that when you make this update, I would greatly appreciate it.

DavidAnson:
You got everything, thank you! There’s just one small request remaining and then I should be able to accept the PR this evening or over the weekend. Thank you for the contribution!

jthomerson:
I renamed both variables to make them consistent with what you mentioned wanting.

PR #262, opened 2020-03-02
Fix NPE with wrapped links

miffels:
The regex change addresses an issue when due to automated line wrapping or other f-ups links get wrapped like this:

```markdown
* This is a badly wrapped [ Link with a leading space
     ](https://acme.com)
```

Before the change, the failed match would result in an NPE:

```
root/node_modules/markdownlint/lib/markdownlint.js:648
      throw error;
      ^

TypeError: Cannot read property 'index' of null
    at root/node_modules/markdownlint/lib/md039.js:32:34
    at Array.forEach (<anonymous>)
    at root/node_modules/markdownlint/lib/md039.js:20:16
    at forToken (root/node_modules/markdownlint/helpers/helpers.js:151:7)
    at Array.forEach (<anonymous>)
    at filterTokens (root/node_modules/markdownlint/helpers/helpers.js:149:17)
    at Object.MD039 [as function] (root/node_modules/markdownlint/lib/md039.js:14:5)
    at forRule (root/node_modules/markdownlint/lib/markdownlint.js:466:20)
    at Array.forEach (<anonymous>)
    at lintContent (root/node_modules/markdownlint/lib/markdownlint.js:512:14)
```

Alternatively, one could add error handling to ensure the regex actually matches:

```javascript
// Before line 32
if(!match) console.error("Could not match " + line");
```

### DIFF
--- a/lib/md039.js
+++ b/lib/md039.js
@@ -4,7 +4,7 @@
 
 const { addErrorContext, filterTokens } = require("../helpers");
 
-const spaceInLinkRe = /\[(?:\s+(?:[^\]]*?)\s*|(?:[^\]]*?)\s+)](?=\(\S*\))/;
+const spaceInLinkRe = /\[|^(?:\s+(?:[^\]]*?)\s*|(?:[^\]]*?)\s+)](?=\(\S*\))/;
 
 module.exports = {
   "names": [ "MD039", "no-space-in-links" ],


miffels:
I notice some of the related (and some unrelated...?) checks fail, so I would appreciate input on how you think this should be handled :)

DavidAnson:
I’ll need to investigate the issue first. I should be back to looking at this project in a few days.

DavidAnson:
After looking into this, I pushed a commit that handles the case when the RegExp match fails (as with wrapped links).

DavidAnson:
Thanks for raising the issue!

PR #290, opened 2020-05-22
chore: Add nightly test of large repos

nschonni:


DavidAnson:
I’m concerned the clone cost of Docs may be too large, even just daily. But maybe not. Could you please start with a PR to `test-repos.yml` to use `clone-large` instead (like you did here)? That will give us a comparison against the current 40s run time today: https://github.com/DavidAnson/markdownlint/actions?query=workflow%3ATestRepos

If it’s pretty close to the 2m time that CI runs now, we can feel good about going with that change instead of a new workflow.

nschonni:
@DavidAnson I through an extra commit in this PR to get the timings you were looking for. I can just rebase it off afterwards

DavidAnson:
Perfect.

DavidAnson:
It’s still faster than CI! Let’s just make it part of the existing action. Thanks for doing this!

nschonni:
OK, I'll rebase out the first commit and fix the commit message

nschonni:
Done, unless you just want to flatten down the "large" with the normal. I'm guessing the CI might be getting some download/cache bump that won't be as quick locally though

DavidAnson:
Right. “Large” takes a while to run on my Raspberry Pi.

PR #311, opened 2020-08-05
Document which rules exist only in Node/Ruby implementations

ChrisBAshton:
I encountered this when trying to configure the 'proper-names' rule:

```
RuntimeError: No such rule: MD044
```

It looks as though several rules are platform-specific, according to:
https://dev.to/jonasbn/blog-post-markdownlint-24ig

Thanks for this project - super useful!

EDIT: have just seen the [CONTRIBUTING.md](https://github.com/DavidAnson/markdownlint/blob/main/CONTRIBUTING.md). I can turn this into an issue if you prefer and we can discuss how best to represent status.

### DIFF
--- a/doc/Rules.md
+++ b/doc/Rules.md
@@ -1581,6 +1581,8 @@ information: <https://cirosantilli.com/markdown-style-guide#top-level-header>.
 
 ## MD042 - No empty links
 
+Availability: Node version only
+
 Tags: links
 
 Aliases: no-empty-links
@@ -1614,6 +1616,8 @@ Rationale: Empty links do not lead anywhere and therefore don't function as link
 <a name="md043"></a>
 
 ## MD043 - Required heading structure
+
+Availability: Node version only
 
 Tags: headings, headers
 
@@ -1681,6 +1685,8 @@ a set of similar content.
 
 ## MD044 - Proper names should have the correct capitalization
 
+Availability: Node version only
+
 Tags: spelling
 
 Aliases: proper-names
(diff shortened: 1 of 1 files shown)


DavidAnson:
The Node implementation documents its rules here: https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md

The Ruby implementation documents its rules here: https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md

Trying to keep these two lists synchronized as both projects develop independently seems like it would be a maintenance challenge. For example, this PR suggests that ￼`MD046 - Code block style` is only available in the Ruby implementation, but that's incorrect because it does so by annotating that rule documentation in the Node repository.

Additionally, there are a handful of rules that behave slightly differently across the two platforms or that support a different set of parameters. Trying to describe this difference in each project's rule list seems like another challenge and something that will cause confusion for people who are just there to learn what the one tool they are using can do.

As such, I am not in favor of making a change like this unless you have a way to address these concerns?￼

waldyrious:
How about a separate document listing only the differences (e.g. like Julia has a page in its manual dedicated to [notable differences from other languages](https://docs.julialang.org/en/v1/manual/noteworthy-differences/index.html))? That way the main rules file would remain clean, but the differences would not be undocumented.

DavidAnson:
@waldyrious That might work! There are probably quite a few topics to cover, but if someone wants to put something like this together, that would be nice. If not, I will add a note to my TODO list.￼

ChrisBAshton:
Thanks both. I'd somehow missed that I was contributing to DavidAnson/markdownlint, and not markdownlint/markdownlint. Needed more coffee yesterday!

Will close this PR.

PR #342, opened 2020-10-17
Use require.resolve as a fallback of path.resolve

kachkaev:
Hi @DavidAnson! We had a [discussion](https://github.com/igorshubovych/markdownlint-cli/issues/97) about resolving `config.extends` in the CLI some time ago; this PR is a follow-up.

My initial question a few months ago was about extending `js` files in an auto-discovered `json` config and you explained some unwanted security implications of a change in the logic. We also discussed an opportunity of using `require.resolve` as a risk-free option for those who want to use a shared markdownlint config and avoid writing something like this in their `.markdownlint.json`:

```js
{
  "extends": "./node_modules/@my-company/markdownlint-config/index.json",
  "first-line-heading": false,
  "no-inline-html": false
}
```

This is what used to stick with till now.

This weekend [I was playing](https://github.com/kachkaev/njt/pull/29) with [Yarn v2 (Berry)](https://github.com/yarnpkg/berry) and realised that the conversion of `"./node_modules/@my-company/markdownlint-config/index.json"` to `"@my-company/markdownlint-config"` became unavoidable. The folder called `node_modules` no longer exists and the real location of all dependency files is now fully controlled by Yarn. Thus, there is no way to refer to a config file in a third-party package without using `require.resolve`.

This PR adds `require.resolve` as a fallback to `path.resolve` in a fully backwards-compatible manner. It should not affect the security model and does not change the error messages in the existing scenarios. The resolution strategy just becomes a little bit more optimistic and can successfully handle more cases than it used to before. So the change in the logic remains unnoticeable for the existing Markdownlint users, while making it possible to use fancier and shorter `"extends"` paths.

I'd be keen to hear what you think about this change and if you're generally happy with the approach, I'm happy to add tests and resolve any other review comments you might have. WDYT?


DavidAnson:
Since then, I have used `require.resolve` in CLI2, so I am more familiar with it. Thanks for bringing this up again! I will have a look at the changes. I'd definitely want to add test cases for the new scenarios. FYI that I'm preparing for a patch release of the library shortly, so would wait to merge this PR until that went out.

lib/markdownlint.js
```diff
@@ -922,11 +955,10 @@ function readConfig(file, parsers, callback) {
       return callback(new Error(message));
     }
     // Extend configuration
-    const configExtends = config.extends;
-    if (configExtends) {
+    if (config.extends) {
+      const resolvedExtends = resolveConfigExtends(file, config.extends);
```
DavidAnson, line 959: Can we call it extendsFile like before? Will give a smaller diff.

lib/markdownlint.js
```diff
@@ -922,11 +955,10 @@ function readConfig(file, parsers, callback) {
       return callback(new Error(message));
     }
     // Extend configuration
-    const configExtends = config.extends;
```
DavidAnson, line 925: Diff would be smaller without this change, too. I think before/after are pretty similar - let's limit this to what's needed?

lib/markdownlint.js
```diff
@@ -971,14 +1003,11 @@ function readConfigSync(file, parsers) {
     throw new Error(message);
   }
   // Extend configuration
```
DavidAnson, line 973: Same comments as other function.

lib/markdownlint.js
```diff
@@ -895,6 +895,39 @@ function parseConfiguration(name, content, parsers) {
   };
 }
 
+/**
+ * Resolve referenced "extends" path ina configuration file
```
DavidAnson, line 899: "ina"

lib/markdownlint.js
```diff
@@ -895,6 +895,39 @@ function parseConfiguration(name, content, parsers) {
...
+ * using path.resolve() and require.resolve() as a fallback.
+ *
+ * @param {string} configFile Configuration file name.
+ * @param {*} referenceFile Referenced path to resolve.
```
DavidAnson, line 903: Add type, string, right?

lib/markdownlint.js
```diff
@@ -895,6 +895,39 @@ function parseConfiguration(name, content, parsers) {
...
+ * @returns {string}
```
DavidAnson, line 904: Add description.

lib/markdownlint.js
```diff
@@ -895,6 +895,39 @@ function parseConfiguration(name, content, parsers) {
...
+ */
+function resolveConfigExtends(configFile, referenceFile) {
+  const resolvedExtendsFile = path.resolve(
+    path.dirname(configFile),
+    referenceFile
+  );
+
+  try {
+    if (fs.statSync(resolvedExtendsFile).isFile()) {
```
DavidAnson, line 913: This is used by sync and async callers, right? Goal is to keep sync calls out of the async path.

lib/markdownlint.js
```diff
@@ -895,6 +895,39 @@ function parseConfiguration(name, content, parsers) {
...
 
+/**
+ * Resolve referenced "extends" path ina configuration file
+ * using path.resolve() and require.resolve() as a fallback.
+ *
+ * @param {string} configFile Configuration file name.
+ * @param {*} referenceFile Referenced path to resolve.
+ * @returns {string}
+ */
+function resolveConfigExtends(configFile, referenceFile) {
```
DavidAnson, line 906: Can we call it "referenceId", maybe - it might not be a file anymore?

lib/markdownlint.js
```diff
@@ -895,6 +895,39 @@ function parseConfiguration(name, content, parsers) {
...
+    }
+  // eslint-disable-next-line unicorn/prefer-optional-catch-binding
+  } catch (error) {
+    // If fs.statSync has thrown, trying require.resolve
+  }
+
+  try {
+    return require.resolve(path.dirname(configFile), referenceFile);
+  // eslint-disable-next-line unicorn/prefer-optional-catch-binding
+  } catch (error) {
```
DavidAnson, line 924: Are these catches to be extra safe, or would both methods throw in normal use? I prefer to let true exceptions bubble out.

DavidAnson (changes requested):
Makes sense! I left some comments/questions. Will do another review after you have a chance to go over them. Thanks!

kachkaev:
Thanks for a quick reply @DavidAnson! I'm happy to address your comments within 12 hours – just going to sleep now 🙂

kachkaev:
Alternatively, feel free to tweak the PR yourself you that’s preferred. I don’t worry much about the authorship, so will be also happy is you close this PR and implement require.resolve fallback yourself – whatever you find easier. Speak soon!

DavidAnson:
No hurry. Get some sleep!

test/markdownlint-test.js
```diff
@@ -1032,6 +1032,44 @@ tape("configMultiple", (test) => {
     });
 });
 
+tape("configMultipleWithRequireResolveBefore", (test) => {
```
kachkaev, line 1035: I could not find alternatives to Jest’s `before()` / `after()` hooks, so followed the idea in https://github.com/substack/tape/issues/59#issuecomment-32809481

lib/markdownlint.js
```diff
@@ -895,6 +895,39 @@ function parseConfiguration(name, content, parsers) {
...
+    }
+  // eslint-disable-next-line unicorn/prefer-optional-catch-binding
+  } catch (error) {
+    // If fs.statSync has thrown, trying require.resolve
+  }
+
+  try {
+    return require.resolve(referenceId);
+
+  /* c8 ignore next 2 */
```
kachkaev, line 922: I had to add this line after facing quite a mystery. Without it, running `npm run ci; npx c8 report --reporter html` _on Node 12_ results in this:

Screenshot 2020-10-18 at 12 43 30

This is a false-positive because both branches are covered. If I add `console.log` in both, they do show up. Other versions though [don’t suffer from this problem](https://github.com/DavidAnson/markdownlint/runs/1270965501). Thoughts?

lib/markdownlint.js
kachkaev: Generally speaking, I agree with the principle. We can duplicate `resolveConfigExtends` so that we have `resolveConfigExtends` and `resolveConfigExtendsSync` like everywhere else, but here are the things to consider:

- `fs.statSync` is super quick, it’s not the same as calling `fs.readFileSync` or `fs.writeFileSync`, which take forever (on CPU scale)
- Node’s `require.resolve` _is_ sync, which in turn depends on some _sync_ fs calls. There can be quite a few of them actually, which involve getting stats different paths, reading `package.json` file and so on.
- Splitting will add more lines to the file, which increases the surface of the bugs
- An async version of the function will make `readConfig` harder to read because we use callbacks and not async / await
- Splitting will require more tests to be added

Happy for you to make a call on this. If a sync nature of `require.resolve` → `fs` is not a sufficient argument for a trade-off I've made, happy to split the function.

lib/markdownlint.js
kachkaev: The key principle I was trying to follow is backwards-compatibility, including the kinds of error messages people get for the unresolved paths. This is what resulted this logic.

If the original `path.resolve` could not throw, `resolveConfigExtends` should not either. It should first try to resolve the reference using `path.resolve` as before _and only if_ the resulting path does not point to an existing _file_ (not folder), try `require.resolve`. By default, `require.resolve` throws, so if we re-throw, the resulting error message ends up being different from previous versions. Thus, we’d have to qualify this PR as a breaking change, which I don’t think is necessary.

That said, I’m open to suggestions 🙂

lib/markdownlint.js
DavidAnson: That's a good point, let's stick with this for now.

lib/markdownlint.js
DavidAnson: I'm thinking of this as a feature and want it to go out as a semver-breaking change. So I'm not worried about the kind of exception changing. Everything that used to work should continue to behave the same - new scenarios can start working - but things that didn't work can not work differently now. :)

lib/markdownlint.js
DavidAnson: C8 only seems to be reliable in the LATEST Node (14). In fact, it already does not run in the CI workflow for this project for Node 10. On another project of mine, it ONLY runs on Node 14. If the behavior you're seeing here is correct on Node 14, we can tweak `nodejs.yml` to do the same here.

test/config/pseudo-package/package.json
```diff
@@ -0,0 +1,3 @@
+{
+  "main": "copy-of-config-child.json"
```
DavidAnson, line 2: Let's just call the file `config-frompackage.json` or something like that.

test/config/config-parent-of-package.json
```diff
@@ -0,0 +1,7 @@
+{
```
DavidAnson, line 1: Maybe call this file `config-packageparent.json`?

test/markdownlint-test.js
DavidAnson: I'm not sure the symlink will work on Windows. And introducing an ordering dependency here will cause problems later when switching to an async runner. Can we move this to a copy+delete in the single test function? I know if it throws an exception it won't clean up all the way, but that's fine with me.

test/markdownlint-test.js
```diff
@@ -1032,6 +1032,44 @@ tape("configMultiple", (test) => {
...
+  test.plan(0);
+  try {
+    fs.symlinkSync(
+      path.resolve(__dirname, "./config/pseudo-package"),
+      path.resolve(__dirname, "../node_modules/pseudo-package"),
```
DavidAnson, line 1040: Could we place this into a `node_modules` under the `test` directory instead? That will keep it out of the "real" one and it should still be searched first, I think?

test/markdownlint-test.js
```diff
@@ -1059,6 +1097,20 @@ tape("configBadChildFile", (test) => {
...
+  test.plan(4);
+  markdownlint.readConfig("./test/config/config-badchildpackage.json",
+    function callback(err, result) {
+      test.ok(err, "Did not get an error for bad child file.");
+      test.ok(err instanceof Error, "Error not instance of Error.");
+      // @ts-ignore
+      test.equal(err.code, "ENOENT",
+        "Error code for bad child file not ENOENT.");
+      test.ok(!result, "Got result for bad child file.");
+      test.end();
```
DavidAnson, line 1110: "file" -> "package"

DavidAnson: For lines 1104/1108/1109

test/markdownlint-test.js
```diff
@@ -1059,6 +1097,20 @@ tape("configBadChildFile", (test) => {
...
 });
 
+tape("configBadChildPackage", (test) => {
+  test.plan(4);
+  markdownlint.readConfig("./test/config/config-badchildpackage.json",
+    function callback(err, result) {
+      test.ok(err, "Did not get an error for bad child file.");
+      test.ok(err instanceof Error, "Error not instance of Error.");
+      // @ts-ignore
+      test.equal(err.code, "ENOENT",
```
DavidAnson, line 1107: In this case of requiring a module, I think `MODULE_NOT_FOUND` is a better code (per discussion above).

DavidAnson (changes requested):
This is looking really good, thank you!

lib/markdownlint.js
kachkaev: IC. How would you implement resolution? There are several ways, each with its own trade-offs. So I'm curious to know your vision, including the priority of the resolutions and the expected feedback to users.

lib/markdownlint.js
```diff
@@ -895,6 +895,39 @@ function parseConfiguration(name, content, parsers) {
...
+    if (fs.statSync(resolvedExtendsFile).isFile()) {
+      return resolvedExtendsFile;
+    }
+  // eslint-disable-next-line unicorn/prefer-optional-catch-binding
+  } catch (error) {
+    // If fs.statSync has thrown, trying require.resolve
+  }
+
+  try {
+    return require.resolve(referenceId);
```
merceyz, line 920:
```suggestion
    return require.resolve(referenceId, { paths: [configFile] });
```

This must be done on behalf of the configFile, otherwise it will be required from `markdownlint` which doesn't declare it as a dependency. Under PnP this will throw an error and under the node_modules layout makes it rely on hoisting to bring `referenceId` and `markdownlint` to the same level.

lib/markdownlint.js
```diff
@@ -895,6 +895,39 @@ function parseConfiguration(name, content, parsers) {
...
+ */
+function resolveConfigExtends(configFile, referenceId) {
+  const configFileDirname = path.dirname(configFile);
+  const resolvedExtendsFile = path.resolve(configFileDirname, referenceId);
+
+  try {
+    if (fs.statSync(resolvedExtendsFile).isFile()) {
+      return resolvedExtendsFile;
+    }
+  // eslint-disable-next-line unicorn/prefer-optional-catch-binding
```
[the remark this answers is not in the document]
kachkaev: `fs.statSync(resolvedExtendsFile)` would not throw for directories. Given that `require.resolve` comes second, we probably want to try it in all cases except when a file is found. The same problem is with `fs.exists` (it returns true for files _and_ folders).

test/markdownlint-test.js
kachkaev: The problem is that this error comes from `fs.readFile` / `fs.readFileSync` in `readConfig`/`readConfigSync`. At that point we don't know if the resolution was done via `path.resolve` or `require.resolve`. Besides, if a config is like `@my-company/my-package/my-file.json` it may be that the package is resolved but the file in the package is non-existent.

test/markdownlint-test.js
kachkaev: What about `@my-company/my-package/my-file.json` (assuming the package exists, but the file within it does not)?

lib/markdownlint.js
kachkaev: Hi @merceyz! 👋

I tried adding that parameter in my earlier experiments, but this broke the currently tested behaviour. I need to double-check.

lib/markdownlint.js
kachkaev: I tweaked `nodejs.yml` as you suggested, CI passed. Thanks! Resoving.

lib/markdownlint.js
kachkaev: It worked! 👍

.github/workflows/nodejs.yml
```diff
@@ -27,8 +27,8 @@ jobs:
     - name: Install Dependencies
       run: npm install --no-package-lock
     - name: Run All Validations
-      if: ${{ matrix.node-version != '10.x' }}
+      if: ${{ matrix.node-version == '14.x' }}
       run: npm run ci
     - name: Run Tests Only
-      if: ${{ matrix.node-version == '10.x' }}
+      if: ${{ matrix.node-version 1= '14.x' }}
```
fisker, line 33:
```suggestion
      if: ${{ matrix.node-version != '14.x' }}
```

lib/markdownlint.js
```diff
@@ -895,6 +895,39 @@ function parseConfiguration(name, content, parsers) {
...
+ * @param {string} configFile Configuration file name.
+ * @param {string} referenceId Referenced path to resolve.
+ * @returns {string} resolved path to file
+ */
+function resolveConfigExtends(configFile, referenceId) {
+  const configFileDirname = path.dirname(configFile);
+  const resolvedExtendsFile = path.resolve(configFileDirname, referenceId);
+
+  try {
+    if (fs.statSync(resolvedExtendsFile).isFile()) {
```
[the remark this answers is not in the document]
kachkaev: I shared some thoughts on `isFile()` in https://github.com/DavidAnson/markdownlint/pull/342#discussion_r507223896. At the moment, the only resulting error code is `ENOENT`, which can be emitted later from `readFile` / `readFileSync`. 

The current idea is that the resolve function never throws (as before). We could change this, but this would require more design work and could possibly make the change more breaking for the existing users / tools.

.github/workflows/nodejs.yml
kachkaev: # 😱

test/markdownlint-test.js
kachkaev: Although there exists a [`test`](https://www.npmjs.com/package/test) package on npm, I renamed as you suggested. I don't think this will ever cause a clash.

kachkaev: Symlink creation should work on Windows too. CI does not seem to complain.

test/markdownlint-test.js
kachkaev: Symlink creation should work on Windows too. CI does not seem to complain.

lib/markdownlint.js
```diff
@@ -895,6 +895,38 @@ function parseConfiguration(name, content, parsers) {
   };
 }
 
+/**
+ * Resolve referenced "extends" path in a configuration file
+ * using path.resolve() and require.resolve() as a fallback.
+ *
+ * @param {string} configFile Configuration file name.
+ * @param {string} referenceId Referenced identifier to resolve.
+ * @returns {string} Resolved path to file
```
DavidAnson, line 904: Nit: End with ".", please

lib/markdownlint.js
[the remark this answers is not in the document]
DavidAnson: Ok

lib/markdownlint.js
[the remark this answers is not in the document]
DavidAnson: Ok

lib/markdownlint.js
DavidAnson: What you have here seems good! I'd prefer to get rid of the ESLint suppression's, I think the `error` parameter is unnecessary?

lib/markdownlint.js
```diff
@@ -895,6 +895,38 @@ function parseConfiguration(name, content, parsers) {
...
+  }
+
+  try {
+    return require.resolve(referenceId, { "paths": [ configFileDirname ] });
+
+  // eslint-disable-next-line unicorn/prefer-optional-catch-binding
+  } catch (error) {
+    // If require.resolve throws, returning resolvedExtendsFile for BC
+  }
+
```
DavidAnson, line 926: Nit: I think in the rest of the file there are not blank lines between sections like this.

test/markdownlint-test.js
DavidAnson: I was not meaning to suggest adding special error handling, just that if the new behavior gives a better error, that is nice!

test/markdownlint-test.js
```diff
@@ -1032,6 +1032,36 @@ tape("configMultiple", (test) => {
     });
 });
 
+tape("configMultipleWithRequireResolve", (test) => {
+  test.plan(2);
+
+  try {
+    fs.symlinkSync(
+      path.resolve(__dirname, "./config/pseudo-package"),
+      path.resolve(__dirname, "../node_modules/test"),
```
DavidAnson, line 1041: I think I wasn't clear. Instead of adding to root/node_modules, can we add to root/test/node_modules? That has the benefit of not touching the "real" one and also showing up in git status if something goes wrong.

test/markdownlint-test.js
DavidAnson: Agreed. But this test is named "...package", so I thought its error messages should say "package" instead of "file".

DavidAnson (changes requested):
Thanks again! Left a couple more minor comments. This is probably the last round since I am getting nit-picky. I may end up tweaking the changes very slightly after merging, but that's better than giving you a hard time.

DavidAnson:
Please also update the PR to be against next branch instead of main.

lib/markdownlint.js
kachkaev: I tried a few things here, but could not get rid of an ESLint error. Having `} catch {` instead of `} catch (error) {` produces:

```
Parsing error: Unexpected token { eslint
```

Adding it back gives:

```
Remove unused catch binding `error`. eslint(unicorn/prefer-optional-catch-binding)
```

(which is what I am suppressing the same way as this is done in a few existing cases)

test/markdownlint-test.js
kachkaev: IC, IC, yeah I misunderstood what you were suggesting. I moved the symlink to `root/test/node_modules` now. What we can do as well is just to commit `root/test/node_modules/pseudo-package` instead and thus avid the symlinking complexity. This, however, whould require some change in `.gitignore` so I decided to not do this without your pre-approval.

kachkaev:
Thanks for another review @DavidAnson! Happy to do address more comments if you want – no rush with the merge. Let’s do it properly 💪

test/markdownlint-test.js
```diff
@@ -1037,6 +1037,35 @@ tape("configMultiple", (test) => {
...
 });
 
+tape("configMultipleWithRequireResolve", (test) => {
+  test.plan(2);
+  const fakeNodeModulesDir = path.resolve(__dirname, "./node_modules");
+  try {
+    fs.mkdirSync(fakeNodeModulesDir);
+    fs.symlinkSync(
+      path.resolve(__dirname, "./config/pseudo-package"),
+      path.resolve(fakeNodeModulesDir, "test"),
```
DavidAnson, line 1047: Could we have the directory name match this test name? One day, maybe there will be another.

DavidAnson (approved):
Looks great! I left one more tiny comment. I'll probably release the patch version in the next couple of days, then can merge this into `next` branch.

Thanks!!

fisker:
I'm the author of `unicorn/prefer-optional-catch-binding`, why we enabled this rule, but used a lot `// eslint-disable-next-line unicorn/prefer-optional-catch-binding`?

DavidAnson:
@fisker That rule is enabled because I agree with it and it looks to be supported by Node version 10+ (as this library is). You may know better, but I'm guessing that I have ESLint slightly misconfigured to be using a language version that does not allow the optional binding? I will look into that and fix it if so.

fisker:
Ah, the `ecmaVersion` is misconfiged, https://github.com/DavidAnson/markdownlint/blob/82cf68023f7dbd2948a65c53fc30482432195de4/.eslintrc.json#L3

I think it should be at least `2019`?

DavidAnson:
@fisker Yes, probably. If there is a good mapping from Node version to this setting, I'd live to know! I will update this tomorrow and remove the unnecessary suppressions. Thanks!

fisker:
This? https://node.green/#ES2019

DavidAnson:
@fisker I like that site and it's what I use, but note that  some of those features are supported by Node 10 and some are not. I guess what I really want is to specify the minimum Node version to ESLint rather than the JavaScript version. No big deal; I will go with 2019 and make the associated updates. Thanks again!

test/markdownlint-test.js
kachkaev: Do you mean renaming `./config/pseudo-package`, `test` or both? Happy to do so, just need a bit more detail. The test is called `configMultipleWithRequireResolve` (camel-case), which we can’t reuse in dir names without changing.

DavidAnson: Sorry, I was thinking something like:  fakeNodeModule/configMultipleWithRequireResolve

kachkaev: `test/configs/fakeNodeModule/configMultipleWithRequireResolve/package.json` plus `test/configs/fakeNodeModule/configMultipleWithRequireResolve/config.json` ?

or

`test/configs/fakeNodeModule/package.json` plus `test/configs/fakeNodeModule/configMultipleWithRequireResolve.json`?

Sorry got confused a bit. 

[If we look at Prettier](https://github.com/prettier/prettier/tree/master/tests_integration/plugins), they just store `node_modules/whatever` right in their repo, so that they don't need to symlink the files. We could to something similar by moving `test/configs/pseudo-package` to `test/node_modules/pseudo-package` or whatever name that would closer match the test name. I would not use `configMultipleWithRequireResolve` because it's camelCase, which is [not allowed for npm package names](https://docs.npmjs.com/files/package.json#name).

kachkaev: Happy for you to push into this PR if that's less time-consuming 🙂

DavidAnson: test/node_modules/pseudo-package sounds great! Agree with adding an exclude line to .gitignore for test/node_modules. Or neither and let me tweak a bit when I come back to this after the patch release. 😀

kachkaev: Done! 👍

DavidAnson: Love it! No copy, no clean-up.

kachkaev:
Thanks again for reviewing and merging this PR @DavidAnson! Looking forward to trying it out in the released version 🚀

kachkaev:
👋 @DavidAnson! What are your thoughts on releasing the next version, which would include this change?

DavidAnson:
I picked out 6 "enhancement" issues to include with the next release. I'm about half way through them now. Once done, I plan to publish!

PR #346, opened 2020-12-05
chore: Remove make-dir-cli

nschonni:


nschonni:
Oops, was looking at the full string and didn't see any hits. Was removing it because there is a NPM audit warning of one of it's dependencies

PR #347, opened 2020-12-05
chore: Replace make-dir-cli with mkdir

nschonni:
There is a minor npm audit issue with the library.
Since there is no recursive creatation needed, regular
mkdir can be used.


DavidAnson:
Great, mkdir did not used to exist on Windows.

PR #349, opened 2020-12-05
chore: Ignore package-lock.json

nschonni:


DavidAnson:
I'd rather not allow this file to sit around invisibly. More info here:￼ https://github.com/DavidAnson/vscode-markdownlint/pull/136#issuecomment-735023819

nschonni:
This keeps it ignored so it doesn't get published in the `.npmignore` file. Alternately, I think on that other PR, there was a version that used the `.npmrc` as an alternate.

DavidAnson:
I like that it shows up as an added file when it exists so that I know to delete it. :)

nschonni:
OK, I went with the `.npmrc` file and ignored it from the publishing. This will not produce the file with a regular `npm install`, but if it get's there somehow, it will still show up in the diff

.npmrc
```diff
@@ -0,0 +1 @@
+package-lock=false
```
DavidAnson, line 1: Please add newline at end.

DavidAnson (changes requested):
Ok, thanks, let's try this. The commit title may need updating?

.npmrc
nschonni: It's there 😉 , I think GitHub diff might not be showing it that clearly. When there is no space, you usually see the "cancel" like symbol in the diff

.npmignore
```diff
@@ -11,3 +11,4 @@ lib-es3
 npm-debug.log
 test
 test-repos
+.npmrc
```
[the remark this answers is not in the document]
nschonni: Done

.npmrc
nschonni: Do you want me to add an EditorConfig in a separate PR to enforce that (except for the test files)?

DavidAnson: No thanks - I'm not sold on editor config.

DavidAnson (approved):

PR #360, opened 2020-12-28
Fix grammar issues in Rules.md

Yash-Singh1:
Multiple Grammar fixes in `docs/Rules.md`

### DIFF
--- a/doc/Rules.md
+++ b/doc/Rules.md
@@ -161,8 +161,8 @@ document:
 ```
 
 The configured list style can be a specific symbol to use (asterisk, plus, dash),
-can require that usage is consistent within the document, or can require that each
-sublist has a consistent symbol that is different from its parent list.
+to ensure that all list styling is consistent, or to ensure that each
+sublist has a consistent symbol that differs from its parent list.
 
 For example, the following is valid for the `sublist` style because the outer-most
 indent uses asterisk, the middle indent uses plus, and the inner-most indent uses dash:
--- a/doc/Rules.md
+++ b/doc/Rules.md
@@ -264,7 +264,7 @@ Some test
 ```
 
 Note: This rule is triggered for the following scenario because the unordered
-sublist is not recognized as such by the parser. Not being nested with 3 characters
+sublist is not recognized as such by the parser. Not being nested 3 characters
 as required by the outer ordered list, it creates a top-level unordered list
 instead.
 
@@ -426,7 +426,7 @@ Aliases: no-reversed-links
 
 Fixable: Most violations can be fixed by tooling
 
-This rule is triggered when a piece of text that appears to be a link is encountered, but
+This rule is triggered when text that appears to be a link is encountered, but
 where the syntax appears to have been reversed (the `[]` and `()` are
 reversed):
 
@@ -514,7 +514,7 @@ line length. This allows you to still include items such as long URLs without
 being forced to break them in the middle. To disable this exception, set the
 `strict` parameter to `true` to report an issue when any line is too long.
 To warn for lines that are too long and could be fixed but allow lines without
-spaces set the `stern` parameter to `true`.
+spaces, set the `stern` parameter to `true`.
 
 For example (assuming normal behavior):
 
@@ -535,7 +535,7 @@ Code blocks are included in this rule by default since it is often a
 requirement for document readability, and tentatively compatible with code
 rules. Still, some languages do not lend themselves to short lines.
 
-Rationale: Extremely long lines can be difficult to work with some editors.
+Rationale: Extremely long lines can be difficult to work with in some editors.
 More information: <https://cirosantilli.com/markdown-style-guide#line-wrapping>.
 
 <a name="md014"></a>
@@ -807,7 +807,7 @@ Aliases: no-duplicate-heading, no-duplicate-header
 
 Parameters: siblings_only, allow_different_nesting (boolean; default `false`)
 
-This rule is triggered if there are multiple headings in the document that has
+This rule is triggered if there are multiple headings in the document that have
 the same text:
 
 ```markdown
@@ -863,9 +863,9 @@ document:
 # Another top-level heading
 ```
 
-To fix, structure your document so that there is a single h1 heading that is
-the title for the document and all later headings are h2 or lower level
-headings:
+To fix, structure your document so there is a single h1 heading that is
+the title for the document. Subsequent headings must be
+lower-level headings (h2, h3, etc.):
 
 ```markdown
 # Title
@@ -917,7 +917,7 @@ To fix this, remove the trailing punctuation:
 
 Note: The `punctuation` parameter can be used to specify what characters count
 as punctuation at the end of a heading. For example, you can change it to
-`".,;:"` to allow headings that end with an exclamation point. A question mark is
+`".,;:"` to allow headings that end with an exclamation point. Question marks is
 allowed by default because of how common it is in headings of FAQ-style documents.
 Setting the `punctuation` parameter to `""` allows all characters - and is
 equivalent to disabling the rule.
--- a/doc/Rules.md
+++ b/doc/Rules.md
@@ -1565,7 +1565,7 @@ correct syntax highlighting for code. More information:
 
 <a name="md041"></a>
 
-## MD041 - The first line in the file should be a top-level heading
+## MD041 - First line in a file should be a top-level heading
 
 Tags: headings, headers
 
--- a/doc/Rules.md
+++ b/doc/Rules.md
@@ -1,7 +1,7 @@
 # Rules
 
 This document contains a description of all rules, what they are checking for,
-as well as an examples of documents that break the rule and corrected
+as well as examples of documents that break the rule and corrected
 versions of the examples. Any rule whose heading is ~~struck through~~ is
 deprecated, but still provided for backward-compatibility.
 
@@ -25,7 +25,7 @@ We skipped out a 2nd level heading in this document
 ```
 
 When using multiple heading levels, nested headings should increase by only one
-level at a time:
+the level at a time:
 
 ```markdown
 # Heading 1
@@ -47,7 +47,7 @@ when skipped - especially for accessibility scenarios. More information:
 
 <a name="md002"></a>
 
-## ~~MD002 - First heading should be a top level heading~~
+## ~~MD002 - First heading should be a top-level heading~~
 
 Tags: headings, headers
 
@@ -59,26 +59,26 @@ Parameters: level (number; default 1)
 > [MD041/first-line-heading](#md041) offers an improved implementation.
 
 This rule is intended to ensure document headings start at the top level and
-is triggered when the first heading in the document isn't a h1 heading:
+is triggered when the first heading in the document isn't an h1 heading:
 
 ```markdown
-## This isn't a H1 heading
+## This isn't an H1 heading
 
 ### Another heading
 ```
 
-The first heading in the document should be a h1 heading:
+The first heading in the document should be an h1 heading:
 
 ```markdown
-# Start with a H1 heading
+# Start with an H1 heading
 
 ## Then use an H2 for subsections
 ```
 
-Note: The `level` parameter can be used to change the top level (ex: to h2) in
-cases where a h1 is added externally.
+Note: The `level` parameter can be used to change the top-level (ex: to h2) in
+cases where an h1 is added externally.
 
-Rationale: The top level heading often acts as the title of a document. More
+Rationale: The top-level heading often acts as the title of a document. More
 information: <https://cirosantilli.com/markdown-style-guide#top-level-header>.
 
 <a name="md003"></a>
@@ -127,7 +127,7 @@ Setext style H2
 
 Note: the configured heading style can be a specific style to use (atx,
 atx_closed, setext, setext_with_atx, setext_with_atx_closed), or simply require
-that the usage be consistent within the document.
+that the usage is consistent within the document.
 
 Rationale: Consistent formatting makes it easier to understand a document.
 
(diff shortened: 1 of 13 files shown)


doc/Rules.md
```diff
@@ -25,7 +25,7 @@ We skipped out a 2nd level heading in this document
 ```
 
 When using multiple heading levels, nested headings should increase by only one
-level at a time:
+the level at a time:
```
waldyrious, line 28: This change does not seem correct.

doc/Rules.md
```diff
@@ -127,7 +127,7 @@ Setext style H2
 
 Note: the configured heading style can be a specific style to use (atx,
 atx_closed, setext, setext_with_atx, setext_with_atx_closed), or simply require
-that the usage be consistent within the document.
+that the usage is consistent within the document.
```
waldyrious, line 130: "require that X be Y" is [correct English](https://ell.stackexchange.com/a/46833/107236).

doc/Rules.md
```diff
@@ -161,8 +161,8 @@ document:
 ```
 
 The configured list style can be a specific symbol to use (asterisk, plus, dash),
-can require that usage be consistent within the document, or can require that each
-sublist have a consistent symbol that is different from its parent list.
+can require that users be consistent within the document, or can require that each
+sublist has a consistent symbol that is different from its parent list.
```
waldyrious, line 165: Same as above about "require that X be Y". This particular change should not be applied.

doc/Rules.md
```diff
@@ -311,7 +311,7 @@ Corrected Example:
 ```
 
 Note: This rule applies to a sublist only if its parent lists are all also
-unordered (otherwise, extra indentation of ordered lists interferes with the
+unordered (otherwise, an extra indentation of ordered lists interferes with the
```
waldyrious, line 314: In this context, "indentation" is uncountable, so adding the "an" is not warranted.

doc/Rules.md
DavidAnson: Agreed, and thanks for helping review!

doc/Rules.md
DavidAnson: Agree, but I think the proposed change may be more understandable.

DavidAnson:
Thanks @Yash-Singh1, I'll do my own review soon - probably after you address comments by @waldyrious. Please update the commit message to something like "Fix grammar issues in Rules.md."

doc/Rules.md
```diff
@@ -426,7 +426,7 @@ Aliases: no-reversed-links
 
 Fixable: Most violations can be fixed by tooling
 
-This rule is triggered when text that appears to be a link is encountered, but
+This rule is triggered when the text that appears to be a link is encountered, but
```
waldyrious, line 429: This change is not actually correct, since this isn't a reference to any specific bit of text, but to text in general.

doc/Rules.md
```diff
@@ -47,7 +47,7 @@ when skipped - especially for accessibility scenarios. More information:
 
 <a name="md002"></a>
 
-## ~~MD002 - First heading should be a top level heading~~
+## ~~MD002 - First heading should be a top-level heading~~
```
DavidAnson, line 50: I'm fine with this change, but please also make this update to the rule description in README.md and md002.js. You may also need to run the update schema script in package.json to update the schema file. (Global search is a good way to confirm.) Thanks!

doc/Rules.md
```diff
@@ -446,7 +446,7 @@ Note: [Markdown Extra](https://en.wikipedia.org/wiki/Markdown_Extra)-style footn
 For (example)[^1]
 ```
 
-Rationale: Reversed links are not rendered as usable links.
+Rationale: Reversed links are not rendered as useful links.
```
waldyrious, line 449: Usable is the correct term here. "Useful" is a subjective assessment, whereas "usable" actually indicates that the target object is functioning.

DavidAnson (changes requested):

DavidAnson:
Also, please update the PR to be against next branch (not main).

doc/Rules.md
```diff
@@ -478,7 +478,7 @@ Some text here
 Some more text here
 ```
 
-Note: this rule will not be triggered if there are multiple consecutive blank
+Note: this rule will not be triggered if there are multiple consecutive blanks
 lines inside code blocks.
```
waldyrious, line 482: Note that the text here originally read "blank lines", so changing to "blanks" makes this incorrect.

doc/Rules.md
```diff
@@ -526,7 +526,7 @@ This-line-is-also-okay-because-there-are-no-spaces
 ```
 
 In `strict` or `stern` modes, the two middle lines above are a violation. The
-third line is a violation in `strict` mode, but allowed in `stern` mode.
+the third line is a violation in `strict` mode but allowed in `stern` mode.
```
waldyrious, line 529: Double "the"?

doc/Rules.md
```diff
@@ -535,7 +535,7 @@ Code blocks are included in this rule by default since it is often a
 requirement for document readability, and tentatively compatible with code
 rules. Still, some languages do not lend themselves to short lines.
 
-Rationale: Extremely long lines can be difficult to work with in some editors.
+Rationale: Extremely long lines can be difficult to work with with some editors.
```
waldyrious, line 538: Double "with"?

doc/Rules.md
```diff
@@ -917,7 +917,7 @@ To fix this, remove the trailing punctuation:
 
 Note: The `punctuation` parameter can be used to specify what characters count
 as punctuation at the end of a heading. For example, you can change it to
-`".,;:"` to allow headings that end with an exclamation point. Question mark is
+`".,;:"` to allow headings that end with an exclamation point. The question mark is
```
waldyrious, line 920: I think this would read better as "A question mark is allowed...".

doc/Rules.md
```diff
@@ -1012,7 +1012,7 @@ Parameters: style ("one", "ordered", "one_or_ordered", "zero"; default "one_or_o
 This rule is triggered for ordered lists that do not either start with '1.' or
 do not have a prefix that increases in numerical order (depending on the
 configured style). The less-common patterns of using '0.' as a first prefix or
-for all prefixes is also supported.
+for all prefixes are also supported.
```
waldyrious, line 1015: The appropriate correction in this sentence would be to change "patterns" to "pattern", rather than to change the verb.

doc/Rules.md
```diff
@@ -1288,7 +1288,7 @@ Note: To allow specific HTML elements, use the 'allowed_elements' parameter.
 
 Rationale: Raw HTML is allowed in markdown, but this rule is included for
 those who want their documents to only include "pure" markdown, or for those
-who are rendering markdown documents in something other than HTML.
+who is rendering markdown documents in something other than HTML?
```
waldyrious, line 1291: "those" is plural, so the verb change is unwarranted. And why the question mark at the end?...

doc/Rules.md
```diff
@@ -1510,7 +1510,7 @@ Rationale: Violations of this rule can lead to improperly rendered content.
 
 <a name="md039"></a>
 
-## MD039 - Spaces inside link text
+## MD039 - Spaces inside the link text
```
waldyrious, line 1513: As mentioned in a previous comment, "text" here doesn't refer to a specific instance, so there's no need to add "the".

doc/Rules.md
```diff
@@ -1679,7 +1679,7 @@ To allow optional headings as with the following structure:
 ```
 
 Use the special value `"*"` meaning "zero or more unspecified headings" or the
-special value `"+"` meaning "one or more unspecified headings" and set the
+the special value `"+"` meaning "one or more unspecified headings" and set the
```
waldyrious, line 1682: Double "the".

doc/Rules.md
```diff
@@ -1733,13 +1733,13 @@ Rationale: Incorrect capitalization of proper names is usually a mistake.
...
 
-## MD045 - Images should have alternate text (alt text)
+## MD045 - Images should have alternate text (alt-text)
 
 Tags: accessibility, images
 
 Aliases: no-alt-text
 
-This rule is triggered when an image is missing alternate text (alt text) information.
+This rule is triggered when an image is missing alternate text (alt-text) information.
```
waldyrious, line 1742: "alt text" (without the hyphen) is the correct spelling.

doc/Rules.md
```diff
@@ -1870,6 +1870,6 @@ document:
 ````
 
 The configured list style can be a specific symbol to use (backtick, tilde), or
-can require that usage be consistent within the document.
+can require that users be consistent within the document.
```
waldyrious, line 1873: Same as above about the subjunctive mood — the original phrasing was correct.

doc/Rules.md
waldyrious: I'm no expert, not a native English speaker, so take this with a grain of salt, but I think "require that X is Y" is actually incorrect. If an alternative phrasing is preferred, I'd suggest "require the usage to be consistent". WDYT?

Yash-Singh1: I just checked on an AI grammar checker and it seems to say that is consistent is better than be consistent.

doc/Rules.md
Yash-Singh1: I changed it "the" -> "a piece of" now.

doc/Rules.md
waldyrious: I hardly would trust the output of an AI grammar checker over the assessment of native human speakers. @DavidAnson any thoughts about my suggestion above?

doc/Rules.md
```diff
@@ -161,8 +161,8 @@ document:
 ```
 
 The configured list style can be a specific symbol to use (asterisk, plus, dash),
-can require that usage be consistent within the document, or can require that each
-sublist have a consistent symbol that is different from its parent list.
+can require that usage is consistent within the document, or can require that each
+sublist has a consistent symbol that is different from its parent list.
```
waldyrious, line 165: I insist on the alternative I suggested [above](https://github.com/DavidAnson/markdownlint/pull/360/files#r549501823): "require each sublist to have a consistent symbol".

doc/Rules.md
```diff
@@ -426,7 +426,7 @@ Aliases: no-reversed-links
 
 Fixable: Most violations can be fixed by tooling
 
-This rule is triggered when text that appears to be a link is encountered, but
+This rule is triggered when a piece of text that appears to be a link is encountered, but
```
waldyrious, line 429: Again, I feel the original phrasing was fine, but happy to defer to @DavidAnson's judgment since this change is not incorrect, just redundant IMO.

doc/Rules.md
```diff
@@ -535,7 +535,7 @@ Code blocks are included in this rule by default since it is often a
 requirement for document readability, and tentatively compatible with code
 rules. Still, some languages do not lend themselves to short lines.
 
-Rationale: Extremely long lines can be difficult to work with in some editors.
+Rationale: Extremely long lines can be difficult to work with some editors.
```
waldyrious, line 538: The "in" should not be removed. The "with" in "work with" does not refer to the editors, but to the long lines, so the "in" is still needed.

doc/Rules.md
```diff
@@ -807,7 +807,7 @@ Aliases: no-duplicate-heading, no-duplicate-header
 
 Parameters: siblings_only, allow_different_nesting (boolean; default `false`)
 
-This rule is triggered if there are multiple headings in the document that have
+This rule is triggered if there are multiple headings in the document that has
```
waldyrious, line 810: The "have" here refers to the headings (plural), not the document (singular). The original phrasing was correct.

doc/Rules.md
```diff
@@ -845,26 +845,26 @@ heading name; headings with the same content can cause problems with that.
...
 ```markdown
 # Top level heading
 
-# Another top level heading
+# Another top-level heading
 ```
 
 To fix, structure your document so that there is a single h1 heading that is
-the title for the document, and all later headings are h2 or lower level
+the title for the document and all later headings are h2 or lower level
```
waldyrious, line 867: IMO, removal of this comma makes the sentence harder to read.

doc/Rules.md
```diff
@@ -1373,7 +1373,7 @@ or match the given style if the rule is so configured:
 ```
 
 Note: by default, this rule is configured to just require that all horizontal
-rules in the document are the same, and will trigger if any of the horizontal
+rules in the document are the same and will trigger if any of the horizontal
```
waldyrious, line 1376: I think the comma should be kept, as it helps pace the reading of this long sentence.

doc/Rules.md
DavidAnson: What about:
> To fix, structure your document so there is a single h1 heading that is the title for the document. Subsequent headings must be lower-level headings h2, h3, etc.:

doc/Rules.md
```diff
@@ -917,7 +917,7 @@ To fix this, remove the trailing punctuation:
 
 Note: The `punctuation` parameter can be used to specify what characters count
 as punctuation at the end of a heading. For example, you can change it to
-`".,;:"` to allow headings that end with an exclamation point. Question mark is
+`".,;:"` to allow headings that end with an exclamation point. A question mark is
```
DavidAnson, line 920: I think I prefer this how it was.

doc/Rules.md
DavidAnson: I like this one better the new way.

DavidAnson (changes requested):
Thanks very much for doing this! Some of the comments are subjective and open to debate, sorry about that. But the majority are a definite improvement. I've tried to ask for changes only where I had a strong feeling (vs. fighting over things that don't matter much).

doc/Rules.md
Yash-Singh1: Compound sentences introducing a new subject need a comma. In this scenario, a new subject is not introduced.

doc/Rules.md
Yash-Singh1: Yep, breaking into two sentences seems better here.

doc/Rules.md
Yash-Singh1: Is the newer one better?:

> to ensure that all list styling is consistent, or to ensure that each
sublist has a consistent symbol that differs from its parent list.

DavidAnson:
I've merged this with 65885fa899a10f5adde58b68e4649fd5d6eadc29, thanks!

DavidAnson:
@Yash-Singh1, I just noticed from your profile that you are a student. Great work, I appreciate the contribution here!

PR #385, opened 2021-04-06
Add spaces_per_tab param to MD010

### LINKED ISSUE #384: Add ability to configure spaces to fix to in no-hard-tabs

Yash-Singh1:
Fixes #384

### DIFF
--- a/lib/md010.js
+++ b/lib/md010.js
@@ -14,10 +14,10 @@ module.exports = {
   "function": function MD010(params, onError) {
     const codeBlocks = params.config.code_blocks;
     const includeCodeBlocks = (codeBlocks === undefined) ? true : !!codeBlocks;
-    const spaceMultiplierProvided = params.config.spaces;
-    const spaceMultiplier = (spaceMultiplierProvided === undefined) ?
+    const spacesPerTab = params.config.spaces_per_tab;
+    const spaceMultiplier = (spacesPerTab === undefined) ?
       1 :
-      Number(spaceMultiplierProvided);
+      Number(spacesPerTab);
     forEachLine(lineMetadata(), (line, lineIndex, inCode) => {
       if (!inCode || includeCodeBlocks) {
         let match = null;
--- a/lib/md010.js
+++ b/lib/md010.js
@@ -17,7 +17,7 @@ module.exports = {
     const spacesPerTab = params.config.spaces_per_tab;
     const spaceMultiplier = (spacesPerTab === undefined) ?
       1 :
-      Number(spacesPerTab);
+      Math.max(0, Number(spacesPerTab));
     forEachLine(lineMetadata(), (line, lineIndex, inCode) => {
       if (!inCode || includeCodeBlocks) {
         let match = null;
--- a/lib/md010.js
+++ b/lib/md010.js
@@ -24,7 +24,6 @@ module.exports = {
         while ((match = tabRe.exec(line)) !== null) {
           const column = match.index + 1;
           const length = match[0].length;
-          const paddingMultiplier = column === 1 ? spaceMultiplier : 1;
           addError(
             onError,
             lineIndex + 1,
@@ -34,7 +33,7 @@ module.exports = {
             {
               "editColumn": column,
               "deleteCount": length,
-              "insertText": "".padEnd(length * paddingMultiplier)
+              "insertText": "".padEnd(length * spaceMultiplier)
             });
         }
       }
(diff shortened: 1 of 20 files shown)


test/markdownlint-test-result-object.js
```diff
@@ -346,7 +346,13 @@ test.cb("resultFormattingV3", (t) => {
         "# Heading   \n" +
         "\n" +
         "Text\ttext\t\ttext\n" +
-        "Text * emphasis * text"
+        "Text * emphasis * text\n" +
+        "\t\tInformation"
```
DavidAnson, line 350: Is there a reason to test it here? If not, it would be better (also easier) to add a new test file and use the "{MD010}" syntax. I'd want one for some number (4?) and another for 0. There should already be a set of files that verify the fix for this rule.

DavidAnson (changes requested):
Thanks! I left a few comments and ideas for review. I like the idea.

test/markdownlint-test-result-object.js
Yash-Singh1: > Is there a reason to test it here?

It was one of the tests on MD010, so I decided to use it.

> If not, it would be better (also easier) to add a new test file

Could you point to some examples? Even though I found multiple test files, I was not able understand where it was being tested.

> use the "{MD010}" syntax

What is this syntax? How would I tell the expected fixes using this syntax?

DavidAnson: Have a look at this set of files:
- https://github.com/DavidAnson/markdownlint/blob/main/test/detailed-results-html-tags.md
- https://github.com/DavidAnson/markdownlint/blob/main/test/detailed-results-html-tags.json
- https://github.com/DavidAnson/markdownlint/blob/main/test/detailed-results-html-tags.results.json
- https://github.com/DavidAnson/markdownlint/blob/main/test/detailed-results-html-tags.md.fixed

I prefer that style over changing code - the test you changed here is mainly about verifying the result object format - not rule behavior and fixes like the above format.

(My comment about "{MD010}" does not apply here, only to tests that do not verify fix output, sorry.)

DavidAnson: Also, let's test for the default (1), a typical value (4) and the edge case (0).

lib/md010.js
```diff
@@ -14,6 +14,10 @@ module.exports = {
   "function": function MD010(params, onError) {
     const codeBlocks = params.config.code_blocks;
     const includeCodeBlocks = (codeBlocks === undefined) ? true : !!codeBlocks;
+    const spacesPerTab = params.config.spaces_per_tab;
+    const spaceMultiplier = (spacesPerTab === undefined) ?
+      1 :
+      Number(spacesPerTab);
```
DavidAnson, line 20: Math.max(0, Number(spacesPerTab))

DavidAnson (changes requested):
Thanks for the updates - looking good! Answered your question and added one new comment.

test/markdownlint-test-result-object.js
Yash-Singh1: 434edc9

test/markdownlint-test-result-object.js
```diff
@@ -346,7 +346,13 @@ test.cb("resultFormattingV3", (t) => {
         "# Heading   \n" +
         "\n" +
         "Text\ttext\t\ttext\n" +
-        "Text * emphasis * text"
+        "Text * emphasis * text\n" +
```
DavidAnson, line 349: Could the changes to this file be reverted by adding a third set of test files that leave the default value of 1?

test/detailed-results-4-tabs-MD010.results.json
```diff
@@ -0,0 +1,29 @@
+[
+  {
+    "errorContext": null,
+    "errorDetail": "Column: 1",
+    "errorRange": [1, 13],
+    "lineNumber": 1,
+    "ruleDescription": "Hard tabs",
+    "ruleInformation": "https://github.com/DavidAnson/markdownlint/blob/v0.23.1/doc/Rules.md#md010",
```
DavidAnson, line 8: Please update the version to "v0.0.0" in these places so they aren't outdated by version upgrades.

DavidAnson (changes requested):
Nice! One test case change and one simple text replacement request remaining. Thank you!

DavidAnson (approved):
Great, thanks! I'll be rebasing this on to `next` branch in a moment.

PR #393, opened 2021-05-13
a typo :)

tallguyjenks:


DavidAnson (changes requested):
Good find! Please fix this here: https://github.com/DavidAnson/markdownlint/blob/f87f9d080002281d96ac32f119d8c99c028a6ddb/schema/build-config-schema.js#L383

Then run the generator script (it's part of CI, so "npm run ci") to update this file and the sibling JSONC one.

Also, please update PR to be for "next" branch per CONTRIBUTING.md.

PR #401, opened 2021-05-26
Fix tiny mistake in markdownlint.yaml

sandersantema:


DavidAnson:
"headers" is the deprecated way to refer to what the specification calls "headings". I think the example is correct as-is, though I see why that's not obvious.

sandersantema:
Ah I see, I was indeed a bit confused by both comments mentioning the same thing but if it's correct it's correct :) I just noticed github seems to have automatically inserted a newline at the end of the file, weird.

PR #415, opened 2021-07-31
docs: add ignore files and folder section

unional:
The `.markdownignore` file is supported but not mentioned in README.


DavidAnson:
This library does nothing to support those files. That implementation is provided by CLI, CLI2, or the VS Code extension (each of which should already have documentation).

unional:
Ar, sure. Sorry. Didn't notice that links to a different repo. https://github.com/igorshubovych/markdownlint-cli/issues/61

Issue #410, opened 2021-07-03
Feature request: Suggest MD030 configuration when Prettier is detected (ex: tab width)


tats-u:
https://github.com/prettier/prettier/issues/5019

Prettier, the most famous Markdown formatter, outputs "formatted" Markdown when tabwidth is not 2:

https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBaABAMwhEAaECABxgEtoBnZUAQwCc6IB3ABXoSpRoBsmaBPKgQBGdGmADWcGAGUi40lADmyGHQCucAnAC2wuABMDhgDI1l6mkrgAxCHR00YZZchA11MXAQAWMHdwA6j6k8BTyYHAyHKGkAG6h-G5gFEIgihRwdDAsYkqOyJg8mQQAVhQAHgBCYpLSMjQ6cCaKcIXFWiDlFTKKStxwAIrqEPDt3CUg8nSZdG6OdBIGzFD4U3SKMIGkBjA+yAAcAAwERIyZgWJEbmdws3FtBACOI-C5xJzuFKhQcIaGazocBepCBuSsBSQRQmnUyOlIqg0sL6A2GozaUI6BBgNGE212+yQABZsWJSNw+gBhCA6SEgO4AVjW6kyABVcZxoZM4poAJJQYywGRgDYkACCApkMH4A3GmQAvvKgA

Input:

```markdown
- foo
```

Output:

```markdown
-   foo
```

This behavior makes sense when item is multi-lined like:

```markdown
-   foo  
    bar  
    baz
-   qux  
    quux
```

May markdownlint disable or suggest disabling MD030 when Prettier is detected and tabwidth = 4 is used.

DavidAnson:
I'm not sure I agree that:

```md
-   foo  
    bar  
    baz
-   qux  
    quux
```

Makes more sense than:

```md
- foo  
  bar  
  baz
- qux  
  quux
```

Both are allowed by the specification, though I think the latter is more common in practice.

Good news is this is already configurable via the parameters for MD030: https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md030---spaces-after-list-markers

I'm not sure how this library would determine that Prettier was going to be used or with what configuration?

[...]

PR #418, opened 2021-08-09
Add MD030 guide for Prettier users in Rules.md

tats-u:
#410

> You are right. How about adding tips for Prettier users in https://github.com/DavidAnson/markdownlint/blob/v0.23.1/doc/Rules.md#md030 and

### DIFF
--- a/doc/Rules.md
+++ b/doc/Rules.md
@@ -1175,8 +1175,11 @@ or
 1.  Bar
 ```
 
+Also, Prettier, the famous code formatter, forces the latter style (always 3 spaces after the markers) if the indent size is 4 spaces.
+
 To fix this, ensure the correct number of spaces are used after the list marker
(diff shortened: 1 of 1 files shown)


tats-u:
I may have to fix the commit message.

DavidAnson (changes requested):
How about a single new section like:

> Note: The default behavior of Prettier [link to its main page] is xyz. To match that behavior in markdownlint, use the following configuration: [json code block for this rule]

DavidAnson:
I've addressed this slightly differently with https://github.com/DavidAnson/markdownlint/commit/580e57ddeec0d59e65b2872a7e70622fb03858df.

tats-u:
It seems to be better. Thank you.

tats-u:
Sorry for late response.

PR #440, opened 2021-10-06
Add docs file on pre-commit

janosh:


DavidAnson:
Please also add a brief note about this to the end of the Examples section. Just a sentence+link, but something so that someone is searching from the README will find it and go here.￼

Issue #150, opened 2018-10-15
New rule proposal: force emphasis style


byzyk:
Emphasis (bold text copy) can be written in two ways: `__bold text__` or `**bold text**`. 

It would be good to force certain emphasis style to keep it consistent.

Please let me know if this could be merged in core repo so I can work on the implementation.

DavidAnson:
This makes sense, thanks. I assume there would also be an option for “consistent“. I’m asking that new rules start externally- more here: https://github.com/DavidAnson/markdownlint/blob/master/CONTRIBUTING.md

byzyk:
@DavidAnson cool, glad to hear that.

I'm going to implement this in our webpack docs first and give you an update on how it went. Gonna keep you posted.

DavidAnson:
Please let me know if the lack of internal helper methods is too annoying: https://github.com/DavidAnson/markdownlint/issues/134

byzyk:
Hey @DavidAnson!

I've come up with a custom rule for emphasis style. It worked pretty well for our case (see [CI build failed](https://travis-ci.org/webpack/webpack.js.org/builds/447713580)).

Do you think this is something that can be merged to core repo? Feel free to dig into [source code](https://github.com/byzyk/markdownlint-rule-emphasis-style/blob/master/rule.js) if you want to to check implementation.

I'd be happy to submit a PR shortly with test cases should you decide to give a green light on this one 🙂

DavidAnson:
Great! I’ll be working on markdownlint again soon and will look into this (and shared.js) then.

DavidAnson:
@byzyk: Still interested in putting your rule implementation into the core? I just accepted another PR for a new rule, and am open to yours as well. :)

byzyk:
Hi @DavidAnson, sorry for a bit of a delay

Yes totally! Is there anything I should note before submitting a PR? I assume you'd expect tests for the rule, right?

DavidAnson:
Cool!  I’ve been changing the coding style of rules slightly, so please have a look in the `next` branch to see what I mean. There are two new rules there (one from a contributor) that may be helpful. I’d ask that you send your PR from that branch and, yes, we will need tests before it ultimately goes in. :) It’s pretty quick and easy to add tests for most rules, so I hope that won’t be a big investment. Thanks for your help!

PS - I’m thinking to release soon, so if you have something in the next few days it can be in this release. If not, that’s totally fine!

byzyk:
[...]

PR #445, opened 2021-10-18
Add rule to verify style of strong

regseb:
Fix https://github.com/DavidAnson/markdownlint/issues/150


DavidAnson:
(In the original issue, there's someone asking for single _/double *, so two rules seems justified.)

PR #447, opened 2021-10-21
Do not trigger MD033 inside multiple liner code spans

### LINKED ISSUE #436: MD009 inside multi line code span
I expected that this markdown passes the linting successfully:

```md
`foo
<bar/>`
```

However, markdownlint produced this warning `MD033/no-inline-html: Inline HTML`

This is not consistent with how single line code spans are parsed. For example, this does not trigger MD003:

```md
`foo<bar/>`
```

mondeja:
Fixes #436

### DIFF
--- a/demo/markdownlint-browser.js
+++ b/demo/markdownlint-browser.js
@@ -3159,7 +3159,6 @@ module.exports = {
         var allowedElements = params.config.allowed_elements;
         allowedElements = Array.isArray(allowedElements) ? allowedElements : [];
         allowedElements = allowedElements.map(function (element) { return element.toLowerCase(); });
-        // compute inline code spans bounds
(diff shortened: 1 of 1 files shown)


DavidAnson:
Thanks for the contribution! I'll update the contributor instructions soon; there are other changes I want to make there already. I have quickly skimmed the changes here and it makes sense. One idea that might be good (I can't tell easily because I'm on my phone) is reusing the inlineCodeSpanRanges function. It seems at first glance to do the same thing as some of the code you've added. Search for it globally or start from this commit to have a look: https://github.com/DavidAnson/markdownlint/commit/7d409263672eabacdd56e473877a9adea7bcabb4

I hope to do a proper review this evening.

PR #465, opened 2021-12-11
Remove Node 10.x logic from CI workflow.

ddzz:
Node 10.x is not included in the matrix configuration in the CI workflow, so this logic can be removed.


bmish (approved):

DavidAnson (changes requested):
Thanks! Please update the PR to be for "next" branch.

DavidAnson (approved):

PR #469, opened 2021-12-19
Fix array indexing for line disable in the presence of frontmatter

adithyabsk:
* And add unit test
^ Not really sure if I set that up correctly

### DIFF
--- a/lib/markdownlint.js
+++ b/lib/markdownlint.js
@@ -404,7 +404,8 @@ function getEnabledRulesPerLineNumber(
       applyEnableDisable(
         action,
         parameter,
-        enabledRulesPerLineNumber[lineNumber + 1] || {}
+        enabledRulesPerLineNumber[
+          lineNumber + 1 + frontMatterLines.length] || {}
       );
     }
   }
--- a/test/front-matter-ignore-line.md
+++ b/test/front-matter-ignore-line.md
@@ -1,0 +1,9 @@
+---
+a: b
+c: d
+e: f
+g: h
+---
+
+<!-- markdownlint-disable-next-line -->
+<iframe width="100%" height="166" scrolling="no" frameborder="no" allow="autoplay" src="http:www.google.com/very+long+url"></iframe>


lib/markdownlint.js
```diff
@@ -404,7 +404,8 @@ function getEnabledRulesPerLineNumber(
       applyEnableDisable(
         action,
         parameter,
-        enabledRulesPerLineNumber[lineNumber + 1] || {}
+        enabledRulesPerLineNumber[
+          lineNumber + 1 + frontMatterLines.length] || {}
```
DavidAnson, line 408: Put this calculation into a const so it can go back to being one line, please.

test/front-matter-ignore-line.md
```diff
@@ -0,0 +1,9 @@
+---
+a: b
+c: d
+e: f
+g: h
+---
+
+<!-- markdownlint-disable-next-line -->
+<iframe width="100%" height="166" scrolling="no" frameborder="no" allow="autoplay" src="http:www.google.com/very+long+url"></iframe>
```
DavidAnson, line 9: This could be any HTML and the suppression just for the "inline-html" rule, I think?

test/front-matter-ignore-line.md
```diff
@@ -0,0 +1,9 @@
+---
```
DavidAnson, line 1: File name: front-matter-with-disable-next-line.md

test/front-matter-ignore-line.md
```diff
@@ -0,0 +1,9 @@
...
+a: b
+c: d
+e: f
+g: h
+---
+
```
DavidAnson, line 7: Please add another kind of violation and suppression below, after a line with a valid reported violation via {MDxxx}, please.

DavidAnson (changes requested):
Good catch! Please update PR to be for "next" branch and fix CI issues. Thanks!

test/front-matter-ignore-line.md
DavidAnson: Specifically, please use a simpler example like "<hr>".

test/front-matter-ignore-line.md
DavidAnson: For example:

Front matter

Disabled violation

Violation

Disabled violation

DavidAnson:
I made the changes I asked for and pushed a commit based on yours here: https://github.com/DavidAnson/markdownlint/commit/5253669495e7e57aebdccd8822c8c2e311c56f1f

Thanks for reporting this!

adithyabsk:
@DavidAnson thanks for getting this merged in! Sorry, I was tied up and couldn't get to it sooner.

DavidAnson:
No worries, you made a great one line fix and I felt like I was asking for a ton of changes.

PR #471, opened 2021-12-24
Add CI badge to README

ddzz:
Quick PR to add a CI badge to the README.


DavidAnson:
Thanks! I used to have one of these, but I got rid of it. In order to be most meaningful, it should really be against the "next" branch where changes happen. However, that runs CI every night and can fail for random Internet weather reasons, so it frequently showed red even though the state is basically always green.

PR #492, opened 2022-01-23
Update yaml config generation

EmilySeville7cfg:
- [x] more restrictions on property values
- [ ] code refactoring


EmilySeville7cfg:
I'll recreate PR according to contributing rules.

DavidAnson:
Thanks. Why use exclusiveMinimum instead of minimum? minimum seems easier to understand.

EmilySeville7cfg:
There is no difference in this case - both ways are simple. ;) But I can use `minimum`.

DavidAnson:
Cool, thanks!

Issue #493, opened 2022-01-23
Enhance JSON schema generation


EmilySeville7cfg:
Tasks:

- sync changes with changes made [here](https://github.com/SchemaStore/schemastore/pull/2040)
- simplify generation code
- try produce JSON schema with smaller size (remove code duplication)

I am going to create PR for this issue.

PR #494, opened 2022-01-23
Update JSON schema

EmilySeville7cfg:
- [x] sync changes with changes made [here](https://github.com/SchemaStore/schemastore/pull/2040)
- [x] simplify generation code
- [x] try produce JSON schema with smaller size (remove code duplication)

### DIFF
--- a/schema/build-config-schema.js
+++ b/schema/build-config-schema.js
@@ -59,6 +59,7 @@ rules.forEach(function forRule(rule) {
         "level": {
           "description": "Heading level",
           "type": "integer",
+          "exclusiveMinimum": 0,
           "default": 1
         }
       };
@@ -101,6 +102,7 @@ rules.forEach(function forRule(rule) {
         "indent": {
           "description": "Spaces for indent",
           "type": "integer",
+          "exclusiveMinimum": 0,
           "default": 2
         },
         "start_indented": {
@@ -109,9 +111,9 @@ rules.forEach(function forRule(rule) {
           "default": false
         },
         "start_indent": {
-          "description":
-            "Spaces for first level indent (when start_indented is set)",
+          "description": "Spaces for first level indent (when start_indented is set)",
           "type": "integer",
+          "exclusiveMinimum": 0,
           "default": 2
         }
       };
@@ -121,6 +123,7 @@ rules.forEach(function forRule(rule) {
         "br_spaces": {
           "description": "Spaces for line break",
           "type": "integer",
+          "exclusiveMinimum": 0,
           "default": 2
         },
         "list_item_empty_lines": {
@@ -144,7 +147,8 @@ rules.forEach(function forRule(rule) {
         },
         "spaces_per_tab": {
           "description": "Number of spaces for each hard tab",
-          "type": "number",
+          "type": "integer",
+          "exclusiveMinimum": 0,
           "default": 1
         }
       };
@@ -154,6 +158,7 @@ rules.forEach(function forRule(rule) {
         "maximum": {
           "description": "Consecutive blank lines",
           "type": "integer",
+          "exclusiveMinimum": 0,
           "default": 1
         }
       };
@@ -163,16 +168,19 @@ rules.forEach(function forRule(rule) {
         "line_length": {
           "description": "Number of characters",
           "type": "integer",
+          "exclusiveMinimum": 0,
           "default": 80
         },
         "heading_line_length": {
           "description": "Number of characters for headings",
           "type": "integer",
+          "exclusiveMinimum": 0,
           "default": 80
         },
         "code_block_line_length": {
           "description": "Number of characters for code blocks",
           "type": "integer",
+          "exclusiveMinimum": 0,
           "default": 80
         },
         "code_blocks": {
@@ -212,11 +220,13 @@ rules.forEach(function forRule(rule) {
         "lines_above": {
           "description": "Blank lines above heading",
           "type": "integer",
+          "exclusiveMinimum": 0,
           "default": 1
         },
         "lines_below": {
           "description": "Blank lines below heading",
           "type": "integer",
+          "exclusiveMinimum": 0,
           "default": 1
         }
       };
@@ -273,21 +283,25 @@ rules.forEach(function forRule(rule) {
         "ul_single": {
           "description": "Spaces for single-line unordered list items",
           "type": "integer",
+          "exclusiveMinimum": 0,
           "default": 1
         },
         "ol_single": {
           "description": "Spaces for single-line ordered list items",
           "type": "integer",
+          "exclusiveMinimum": 0,
           "default": 1
         },
         "ul_multi": {
           "description": "Spaces for multi-line unordered list items",
           "type": "integer",
+          "exclusiveMinimum": 0,
           "default": 1
         },
         "ol_multi": {
           "description": "Spaces for multi-line ordered list items",
           "type": "integer",
+          "exclusiveMinimum": 0,
           "default": 1
         }
       };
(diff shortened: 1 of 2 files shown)


EmilySeville7cfg:
Closes #493

EmilySeville7cfg:
closes #493

schema/build-config-schema.js
```diff
@@ -59,6 +59,7 @@ rules.forEach(function forRule(rule) {
         "level": {
           "description": "Heading level",
           "type": "integer",
+          "minimum": 1,
```
DavidAnson, line 62: Let's add maximum:6 while we're at it.

schema/build-config-schema.js
```diff
@@ -328,6 +342,7 @@ rules.forEach(function forRule(rule) {
...
```
DavidAnson, line 345: maximum:6 again here, please

schema/build-config-schema.js
```diff
@@ -109,9 +111,9 @@ rules.forEach(function forRule(rule) {
           "default": false
         },
         "start_indent": {
-          "description":
-            "Spaces for first level indent (when start_indented is set)",
```
DavidAnson, line 113: This change introduces a linting error

schema/build-config-schema.js
```diff
@@ -436,11 +451,16 @@ rules.forEach(function forRule(rule) {
   }
   if (custom) {
     // @ts-ignore
-    scheme.type = [ "boolean", "object" ];
+    scheme.type = ["boolean", "object"];
```
DavidAnson, line 454: Linting error here

schema/build-config-schema.js
```diff
@@ -436,11 +451,16 @@ rules.forEach(function forRule(rule) {
...
     scheme.additionalProperties = false;
   }
-  rule.names.forEach(function forName(name) {
-    schema.properties[name] = scheme;
+  rule.names.forEach(function forName(name, index) {
+    if (index == 0)
```
DavidAnson, line 458: The change makes sense, but is hard to read in the GitHub diff due to linting issues - please fix those and I'll have a closer look.

schema/build-config-schema.js
```diff
@@ -436,11 +451,16 @@ rules.forEach(function forRule(rule) {
...
+      schema.properties[name] = scheme;
+    else
+    schema.properties[name] = {
+      "$ref": `#/properties/${rule.names[0]}`
```
DavidAnson, line 462: I love this change IFF VS Code handles it properly - do you happen to know? Otherwise, I'd consider this a regression.

DavidAnson (changes requested):
Makes sense, I love how it reduces redundant redundancy! I left a couple of comments around introducing "maximum". These changes introduce some linting violations that break CI, so please address those. `npm run lint` is a quick way to lint, but you can `npm run ci` locally to see exactly what the PR checks are going to do. Thanks for the contribution!

```
Error:   114:1   error  This line has a length of 86. Maximum allowed is 80  max-len
138
Error:   454:19  error  A space is required after '['                        array-bracket-spacing
139
Error:   454:39  error  A space is required before ']'                       array-bracket-spacing
140
Error:   458:15  error  Expected '===' and instead saw '=='                  eqeqeq
141
Error:   459:7   error  Expected { after 'if' condition                      curly
142
Error:   459:7   error  Expected no linebreak before this statement          nonblock-statement-body-position
143
Error:   461:1   error  Expected indentation of 6 spaces but found 4         indent
144
Error:   461:5   error  Expected { after 'else'                              curly
145
Error:   461:5   error  Expected no linebreak before this statement          nonblock-statement-body-position
146
Error:   462:1   error  Expected indentation of 8 spaces but found 6         indent
147
Error:   463:1   error  Expected indentation of 6 spaces but found 4         indent
```

schema/build-config-schema.js
EmilySeville7cfg: Yes, it's is a expected VS Code feature. :)

schema/build-config-schema.js
```diff
@@ -343,15 +361,17 @@ rules.forEach(function forRule(rule) {
           "description": "List of headings",
           "type": "array",
           "items": {
-            "type": "string"
+            "type": "string",
+            "pattern": "^(\\*|\\+|#{1,6} .*)$"
```
DavidAnson, line 365: Good idea! For ` .*` (includes space), could it be `\\s\\S.*` instead? In other words, single space then non-space then anything.

schema/build-config-schema.js
```diff
@@ -439,8 +459,14 @@ rules.forEach(function forRule(rule) {
     scheme.type = [ "boolean", "object" ];
     scheme.additionalProperties = false;
   }
-  rule.names.forEach(function forName(name) {
-    schema.properties[name] = scheme;
+  rule.names.forEach(function forName(name, index) {
+    if (index === 0) {
```
DavidAnson, line 463: Could we use ternary here?
a = (b) ? c : { ref:... };

DavidAnson (changes requested):
I left one comment about the new pattern and one about simplifying code. I'll want to see ref work in VS Code (I can do this on my machine after changes are done), but happy to hear it is supported!

DavidAnson:
One more question: is it possible to use a relative path for the references? For the VS Code extension, this schema gets embedded in another JSON file, so absolute paths will be wrong. Maybe Code handles that or if not, I can probably rewrite them during import. But if it's possible to use relative paths, that would be more flexible and avoid such problems.

schema/build-config-schema.js
EmilySeville7cfg: I am not sure whether it's a good idea according to [this note](https://json-schema.org/understanding-json-schema/reference/regular_expressions.html):

> The regular expression syntax used is from JavaScript (ECMA 262, specifically). However, that complete syntax is not widely supported, therefore it is recommended that you stick to the subset of that syntax described below.

EmilySeville7cfg:
> One more question: is it possible to use a relative path for the references? For the VS Code extension, this schema gets embedded in another JSON file, so absolute paths will be wrong. Maybe Code handles that or if not, I can probably rewrite them during import. But if it's possible to use relative paths, that would be more flexible and avoid such problems.

I [tried](https://github.com/EmilySeville7cfg/EmilySeville7cfg.github.io/blob/feature/10/schemas/organizations.json#L18) yesterday use relative refs. In VS Code it seems to work perfectly (suggestions are shows as expected) but `ajv-cli` tool [fails](https://github.com/ajv-validator/ajv-cli/issues/201) to check such schemas. Maybe I am mistaken anywhere...

schema/build-config-schema.js
```diff
@@ -439,8 +459,10 @@ rules.forEach(function forRule(rule) {
     scheme.type = [ "boolean", "object" ];
     scheme.additionalProperties = false;
   }
-  rule.names.forEach(function forName(name) {
-    schema.properties[name] = scheme;
+  rule.names.forEach(function forName(name, index) {
+    schema.properties[name] = index === 0 ? scheme : {
```
DavidAnson, line 463:
```suggestion
    schema.properties[name] = (index === 0) ? scheme : {
```

DavidAnson (changes requested):
Looks good - one tiny change I'm trying to add as a suggestion.

schema/build-config-schema.js
DavidAnson: Seems I don't have write permission.

DavidAnson:
I've confirmed the new references break VS Code behavior for `package.json/contributes/configuration/properties/markdownlint.config` (i.e., sub-properties are not properly suggested). This is an improvement anyway and I've made a note to expand references when importing to the VS Code extension.

DavidAnson (approved):
I'm going to manually merge this so I can make the tweak I wanted - thanks for the improvements!

DavidAnson:
Corresponding commit: https://github.com/DavidAnson/markdownlint/commit/245791f79f1dd2f3f983bdabfddfedd7717eecb4

Issue #121, opened 2018-05-23
Rule to validate links via https://github.com/tcort/markdown-link-check/


nschonni:
I've been using https://github.com/tcort/markdown-link-check/ as a separate check to validate links in markdown. I was starting to see if I could leverage the `.markdownlint.json` as a config format for that project too and realized maybe it might make sense to just add that functionality as a rule in this project.
Because it can go out and create network requests, I think having it off by default would make sense.
I may just do it as a custom rule in the end, but I figured I'd check to see if there is interest in this as a core rule.

Example config schema:
```json
    "valid-links": {
      "description": "MD###/valid-links - Validate links",
      "type": [
        "boolean",
        "object"
      ],
      "default": true,
      "properties": {
        "link_types": {
          "description": "Types of links to validate",
          "type": "string",
          "enum": [
            "all",
            "internal",
            "external"
          ],
          "default": "internal"
        },
        "ignore_patterns": {
          "description": "RegExp for ignoring links",
          "type": "array",
          "items": {
            "type": "string"
          },
          "default": []
        },
        "http_headers": {
          "description": "Url and HTTP header pairs when URLs need additional parameters",
          "type": "array",
          "items": {
            "type": "object"
          },
          "default": []
      },
      "additionalProperties": false
```

/cc @tcort @timmkrause

DavidAnson:
Love the idea! This comes up from time to time, so definitely worth thinking about. I think it would be pretty straightforward to implement as a custom rule, especially because the parser will have already pulled out each of the URLs.

What I’d like to do is start everything as a custom rule, see how it does, then look at pulling successful ones into the core library. (With the caveat that I try hard to avoid new dependencies, so maybe attempt this using only the core Node.js modules.)

If you prototype something here, I'd be happy to answer any questions that come up!

(See also https://github.com/DavidAnson/check-pages for a project of mine that checks links.)

nschonni:
Sounds good, I'll see if I can figure out setting up a separate rule repo.
Any advice on naming conventions and namespacing for rules, just to avoid any future issues if you decide it would fit in core?

DavidAnson:
Don’t assign an MDxxx name, just use a name like validate-links and it should be possible to pull it in pretty much as-is. Some other guidance here: https://github.com/DavidAnson/markdownlint/blob/master/doc/CustomRules.md

amimas:
[...]

PR #495, opened 2022-01-25
feat(rules): add `valid-link-fragments`

### LINKED ISSUE #253: Warn against dead links within the document
It would be fantastic if bad anchor-style links were detected and warned.
I wouldn't expect bad external URL links to be caught.

Divlo:
Hey! :wave:

This PR adds a new built-in rule: `valid-link-fragments` to ensure that link fragments are valid within the document.

Even if you recommend first implementing it as a custom rule as an npm package, I think that we could instead disable the rule by default and let the user enable it in their config, and if successful, we enable it by default (in the current implementation of the PR, the rule is enabled by default).

I left some comments to discuss the "know issues" and things that could be implemented in this PR.

fix #253


DavidAnson:
I'll review this soon, thanks! Please have a look at CONTRIBUTING.md in the meanwhile; PR for `next` branch.

DavidAnson:
`fs.existsSync` is a problem. First because it is synchronous and asynchronous callers are supported. Second because it uses `require("fs")` (context: https://github.com/DavidAnson/markdownlint#optionsfs).

lib/md051.js
```diff
@@ -0,0 +1,36 @@
...
+  "function": function MD051(params, onError) {
+    filterTokens(params, "inline", (token) => {
+      token.children.forEach((child) => {
+        const { lineNumber, type, attrs } = child;
+        if (type === "link_open") {
+          attrs.forEach((attr) => {
+            if (attr[0] === "href") {
+              const href = attr[1];
+              const url = new URL(href, pathToFileURL(params.name));
+              url.hash = "";
```
Divlo, line 23: Currently, "we don't care" about the hash, we only check if the file exists but we could do beyond that and also check if the file is a markdown file and if we have a `href` with a hash (e.g: `./README.md#about`), we could check that we actually have a heading with this hash (e.g: `about`).
	
This would fix this issue: https://github.com/DavidAnson/markdownlint/issues/253
I will try to fix that issue in this PR (not yet implemented).

### Potential Implementation

Taking the example of this `href`:  `./README.md#about`
We could read the file content `./README.md`, parse all the headings in an array following the algorithm as specified in this comment: https://github.com/DavidAnson/markdownlint/issues/253#issuecomment-651369616 and then check if this array contains the hash `about`?
Is this would be a viable solution? Simply ask for guidance and open the discussion. :smile:

lib/md051.js
```diff
@@ -0,0 +1,36 @@
...
+              const isRelative = href.startsWith("./") ||
```
Divlo, line 24: Relative links could also be without `./`, for example, `README.md` is a valid relative `href` but currently we don't take this into account to stay simple, the problem is that we don't want to consider for example `google.com` as a relative link.

Is this implementation viable? Do you have ideas about this?
I think we could only consider relative links as being `href` starting with `./` or `../`, it could be enough for this rule.

lib/md051.js
```diff
@@ -0,0 +1,36 @@
...
+
+module.exports = {
+  "names": [ "MD051", "no-dead-relative-links" ],
+  "description": "No dead relative links",
+  "tags": [ "links" ],
+  "function": function MD051(params, onError) {
+    filterTokens(params, "inline", (token) => {
+      token.children.forEach((child) => {
+        const { lineNumber, type, attrs } = child;
+        if (type === "link_open") {
```
Divlo, line 18: Currently, it only takes into account the attribute `href` of `a` (e.g: `[README](./README.md)`) but we could also check the `src` of `img`, what do you think?

What would be needed there, to check that it is `link_open` or an image?

Divlo (commented):
Left some comments to open the discussion about the implementation.

Thank you very much for taking the time to review this and to consider this PR. :smile:

Divlo:
> I'll review this soon, thanks! Please have a look at CONTRIBUTING.md in the meanwhile; PR for `next` branch.

Oops, yep forgot to target the PR for `next` branch. :smile:

The CI is failing only because of TypeScript types, did I miss something, how can we fix this?

Divlo:
> `fs.existsSync` is a problem. First because it is synchronous and asynchronous callers are supported. Second because it uses `require("fs")` (context: [DavidAnson/markdownlint#optionsfs](https://github.com/DavidAnson/markdownlint#optionsfs)).

Right, it makes sense. :+1:
How are we supposed to use `fs` in the rule code, I don't see how can we access `options.fs`.

We could implement this rule as `"asynchronous": true` and have a function like this:
```js
const isExistingFile = async (path) => {
  try {
    await fs.promises.access(path, fs.constants.F_OK)
    return true
  } catch {
    return false
  }
}
```

DavidAnson:
There is a lot to this, so here are my thoughts in no order in a numbered list so it's easier to refer to them:

1. The built-in rules are all synchronous and completely local which allows them to be very fast which can be important in scenarios like the VS Code extension where they are run on every keypress. This is one reason I've avoided implementing many URL rules so far.
2. Some scenarios like when run in the browser don't really have a concept of file system at all.
3. Incorporating the idea to also read the content of the relative file or validate an HTTP link would make preserving the current performance even more challenging.
4. I think the intent of issue #253 is to detect broken links within the same document only. For that scenario, no file access is needed and this rule could be implemented without issue.
5. Asynchronous rules can only run in an asynchronous context. However, the library and built-in rules need to be able to run in both synchronous and asynchronous context. Therefore, a rule like what is proposed would need to be able to detect the current context and use the appropriate file system call. This information is not currently available. Nor is the "fs" instance exposed.
6. Exposing the file system interface to third-party extensions would require documenting it and ensuring there are enough functions there to be useful. Right now it is private and as minimal as possible. This would be a considerable amount of work.
7. Having the rule fail in some cases because of how it was run is not something I'd like to do. Any built-in rule should work the same however the tool is used.
8. Validating image references seems like a different rule entirely.

Number three above is the only path forward I see right now that doesn't have a lot of open questions/complexity. I think that is a useful rule on its own, though it is more limited than what you envision.

Divlo:
@DavidAnson Some thoughts point by point :smile: :

1\. Indeed `markdownlint` should be fast! :+1:

2\. So we can't use node built-in modules if we need to support browser and Node.js for `markdownlint`?

3\. Yep, same as 1. but in #121, I don't see any objection because of performance checking external URLs for example.

4\. Updated the PR to add this rule instead, so this PR will now fix #253 but yes it is a lot less strict than what I imagined for the rule `no-dead-relative-links` but that would still be great to have at least.

5\. and 6\. Is there is any way to do a custom rule that accesses the file system currently? Basically, #121  is not fixable currently ?

7\. What do you mean that the rule depends on how it was run, is the `no-dead-relative-links` depends on how the tool was run? Could you give example of when it could fail or succeed depending on how it was run?

8\. Yeah I agree, but the implementation would be pretty much the same, the exception that we loop through `img` instead of `a` and that we check `src` instead of `href`.

Thanks for your fast and detailed reply. :smile:

DavidAnson:
1. EMPTY
2. It's probably best not to require Node modules. Some of the core modules like path are available as separate packages that are mostly compatible in the browser, but not all of them. This policy applies primarily to built-in rules; someone writing a custom rule could decide not to support the browser scenario and that would be perfectly fine for them and their users.
3. Attempting to load the content of a URL is arguably worse/slower than hitting the file system because network delays are especially unpredictable and can be very long at times.
4. I appreciate that, thank you. I'll probably look over the code tomorrow. The quick look I already did was good and it seemed like you'd already done most of the things I would ask for.
5. Kind of like I said earlier, a custom rule could decide to take a dependency on the file system API and could decide to be asynchronous (which means it is only supported in asynchronous scenarios). This may be perfectly acceptable for the right audience, but built-in rules are held to a higher standard.
6. See 5
7. The markdownlint library may be invoked by its caller from a synchronous or an asynchronous API. It adjusts its behavior according to that, making any file-system accesses using the corresponding mode. Similarly, asynchronous custom rules will only be run in an asynchronous context. A fully synchronous rule doesn't need to know about this or do anything to handle it; it will be called and behave the same both ways. See also https://github.com/DavidAnson/markdownlint/blob/main/doc/CustomRules.md#asynchronous-rules
8. See my comment above about making network calls. :)

Artem41185 (approved):

README.md
```diff
@@ -106,6 +106,7 @@ playground for learning and exploring.
 * **[MD048](doc/Rules.md#md048)** *code-fence-style* - Code fence style
 * **[MD049](doc/Rules.md#md049)** *emphasis-style* - Emphasis style should be consistent
 * **[MD050](doc/Rules.md#md050)** *strong-style* - Strong style should be consistent
+* **[MD051](doc/Rules.md#md051)** *no-dead-hash-links-within-document* - No dead hash links within document
```
DavidAnson, line 109: How about just calling the rule `valid-link-fragments`?
https://spec.commonmark.org/0.30/#example-500

DavidAnson: "Link fragments should be valid"

doc/Rules.md
```diff
@@ -56,7 +56,8 @@ Aliases: first-heading-h1, first-header-h1
 Parameters: level (number; default 1)
 
 > Note: *MD002 has been deprecated and is disabled by default.*
-> [MD041/first-line-heading](#md041) offers an improved implementation.
+> [MD041/first-line-heading](#md041---first-line-in-a-file-should-be-a-top-level-heading)
```
DavidAnson, line 59: Please keep the short form like it was here and elsewhere.

doc/Rules.md
```diff
@@ -1984,3 +1986,20 @@ The configured strong style can be a specific symbol to use ("asterisk",
 "underscore"), or can require that usage be consistent within the document.
 
 Rationale: Consistent formatting makes it easier to understand a document.
+
```
DavidAnson, line 1989: Please add `<a name="md051"></a>` for linking to this.

doc/Rules.md
```diff
@@ -1984,3 +1986,20 @@ The configured strong style can be a specific symbol to use ("asterisk",
...
+## MD051 - No dead hash links within document
+
+Tags: links
+
+Aliases: no-dead-hash-links-within-document
+
+This rule is triggered whenever a dead hash link is found within the document:
```
DavidAnson, line 1996: This rule is triggered if a link fragment does not correspond to an heading within the document.

DavidAnson: It's probably necessary to say how a heading is turned into a link. It's not part of the CommonMark specification and it may be GitHub-specific. Let's link to GitHub documentation if so.

DavidAnson: Avoid the word "dead" and "slug" if possible, please.

doc/Rules.md
```diff
@@ -1984,3 +1986,20 @@ The configured strong style can be a specific symbol to use ("asterisk",
...
+
+```markdown
+# Title
+
+[Valid link 1](#invalid-title)
```
DavidAnson, line 2001: No need for the number 1 here.

DavidAnson: Let's call it `#invalid-fragment`.

doc/Rules.md
```diff
@@ -1984,3 +1986,20 @@ The configured strong style can be a specific symbol to use ("asterisk",
...
+```
+
+To fix this issue, ensure that the heading with the slug `invalid-title` exists,
+here you could replace `#invalid-title` by `#title`.
```
DavidAnson, line 2005: "... exists (in this case `title`)."

lib/md051.js
```diff
@@ -0,0 +1,50 @@
+// @ts-check
+
+"use strict";
+
+const { filterTokens, addError, forEachHeading } = require("../helpers");
+
+/**
+ * Slugifies strings.
```
DavidAnson, line 8: Something like "Converts a Markdown heading into an HTML fragment according to the rules used by GitHub", maybe.

lib/md051.js
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @param {string} string The string to slugify.
+ * @returns {string} The slugified string.
+ */
+function slugify(string) {
+  return string
+    .toLowerCase()
+    .replace(/[.,/#!$%^&*;:{}[\]=_`~()]/g, "")
+    .split(" ")
```
DavidAnson, line 17: `.replace(/ /g, "-")`?

lib/md051.js
```diff
@@ -0,0 +1,50 @@
...
+}
+
+module.exports = {
+  "names": [ "MD051", "no-dead-hash-links-within-document" ],
+  "description": "No dead hash links within document",
+  "tags": [ "links" ],
+  "function": function MD051(params, onError) {
+    const headings = [];
+    forEachHeading(params, (_heading, content) => {
+      headings.push("#" + slugify(content));
```
DavidAnson, line 28: Make the `#` part of the function?

lib/md051.js
```diff
@@ -0,0 +1,50 @@
...
+    });
+    filterTokens(params, "inline", (token) => {
```
DavidAnson, line 30: Use `forEachInlineChild` instead, I think?

lib/md051.js
```diff
@@ -0,0 +1,50 @@
...
+      token.children.forEach((child) => {
+        const { lineNumber, type, attrs } = child;
+        if (type === "link_open") {
+          const href = attrs.find((attr) => attr[0] === "href");
+          if (href !== undefined &&
```
DavidAnson, line 35: Just `href`

lib/md051.js
```diff
@@ -0,0 +1,50 @@
...
+              typeof href[1] === "string" &&
```
DavidAnson, line 36: Is this necessary? What else would it be?

lib/md051.js
```diff
@@ -0,0 +1,50 @@
...
+              href[1].length >= 2 &&
```
DavidAnson, line 37: Not necessary?

lib/md051.js
```diff
@@ -0,0 +1,50 @@
...
+              href[1].startsWith("#")
+          ) {
+            const hasHeading = headings.includes(href[1]);
```
DavidAnson, line 40: Put this into the `if` since it's only used once.

test/detailed-results-MD041-MD050.md.fixed
```diff
@@ -14,7 +14,7 @@ that capitalizes the
 name wrong twice:
 markdownlint.
 
-A [normal](link) and an [empty one]() and a [fragment](#one).
+A [normal](link) and an [empty one]().
```
DavidAnson, line 17: Please add new `detailed-results-MD051-MD060.results.*` and a scenario for the new rule.

test/no-dead-hash-links-within-document.md
```diff
@@ -0,0 +1,17 @@
+# Valid/Invalid Hash Links within document
```
DavidAnson, line 1: Title case for headings, please

test/no-dead-hash-links-within-document.md
```diff
@@ -0,0 +1,17 @@
...
+
+[Valid link](#validinvalid-hash-links-within-document)
+
+[Valid link](#valid-links)
+
+[Valid link](#valid-h3-heading)
+
+### Valid h3 heading
+
+content...
```
DavidAnson, line 13: `Text` or `Text text text` for consistency.

test/no-dead-hash-links-within-document.md
```diff
@@ -0,0 +1,17 @@
...
+
+## Invalid links
+
+[Invalid link](#dead-slug-no-heading) {MD051}
```
DavidAnson, line 17: Please add a few more test cases. Maybe some that verify symbols get replaced. Let's be sure ATX headings are handled correctly.

test/spaces_inside_codespan_elements.md
```diff
@@ -78,21 +78,21 @@ Text [link](https://example.com/link`link`link`link) text `code`.
 
 Text [link](https://example.com/link "title`title") text `code`.
 
-Text [link](#link`link) text `code`.
+Text [link](#link`link) text `code`. {MD051}
```
DavidAnson, line 81: OR disable MD051 on line 80.

DavidAnson (changes requested):
Made a first pass review. The CI failures look to be due uncommitted changes. Please also update `break-all-the-rules.md` with an instance of MD051. There are more than 1000 failures in `test-repos` ([link](https://github.com/DavidAnson/markdownlint/runs/4974196400?check_suite_focus=true)) - that shouldn't be the case for an existing repository. There may be a bug OR there may be something the rule is not handling well today. Thanks - I'll have a second look when you're ready!

Divlo:
@DavidAnson Thanks for your review! :smile:

I addressed some of your comments (I'll let you "Resolve conversation" where you think it has been addressed), there is still work to be done, see my replies to some of your comments to inform you of the current issues.

doc/Rules.md
Divlo: I changed that because it's not respecting the new rule `valid-link-fragments`.
Currently, we are checking headings, not `id`, not `name` attribute from a link `<a>`.
So yes even if `#md041` is a valid fragment because of `<a name="md041"></a>`, there are no heading with `md041` as content so the rule fails.

I might also get the `id` and `name` attribute to make it working.

lib/md051.js
Divlo: Indeed it is not necessary. My mistake, I would rather work with TypeScript ah ah.

doc/Rules.md
DavidAnson: Ah, yes, this is a valid scenario, so let's allow it.

test/no-dead-hash-links-within-document.md
DavidAnson: I see some new cases, thanks. Please include the two other heading kinds as well:

```markdown
Heading
=======

# Heading #
```

doc/Rules.md
```diff
@@ -1984,3 +1986,26 @@ The configured strong style can be a specific symbol to use ("asterisk",
...
+
+<a name="md051"></a>
+
+## MD051 - Link fragments should be valid
+
+Tags: links
+
+Aliases: valid-link-fragments
+
+This rule is triggered if a link fragment does not correspond to an
```
DavidAnson, line 1998: "an"->"a"

DavidAnson:
I resolved the conversations that looked done, thanks! There are a few left. I left one new comment and commented on your suggestion to include a/name. There's still a CI failure - `npm run ci` will surface it - seems the generated schema is out of date.

I had a look at each of the `mochajs-mocha` violations you show above - these are all valid and a great example of the usefulness of this rule! FYI, `npm run clone-test-repos-large` then `npm run lint-test-repos` will do that analysis locally.

The 3 `mkdocs/mkdocs` issues are valid as well. I sampled a few in `pi-hole/docs` and they were valid as well.

But I can't believe there are that many issues in live docs. I wonder if we could run one of the existing tools to do this analysis and make sure the results line up with what we're seeing here (once `a/name` is handled).

Divlo:
This is a harder PR than I thought. :sweat_smile: 
This is a kind of PR that is difficult to get right first try ah ah, but after all, it's like that we're learning.

> these are all valid and a great example of the usefulness of this rule!

Exactly! :smile: Human makes mistakes, we can make tools like `markdownlint` to make lives easier and avoid these little mistakes.

> But I can't believe there are that many issues in live docs.

I can't believe it too, it just seems "too much".

doc/Rules.md
Divlo: Yes, we should allow it.
I tried but could not figure out, how can I loop trough each inline HTML and check if it's a link, so we add the `name` content inside the `validFragments` variable.

Is there a utility function to do that?

DavidAnson: I don't think so. It might be necessary to use a RegExp which is unfortunate because that's trickier and less reliable.

lib/md051.js
```diff
@@ -0,0 +1,46 @@
...
+ * Converts a Markdown heading into an HTML fragment
+ * according to the rules used by GitHub.
+ *
+ * @param {string} string The string to convert.
+ * @returns {string} The converted string.
+ */
+function convertHeadingToHTMLFragment(string) {
+  return "#" + string
+    .toLowerCase()
+    .replace(/ /g, "-")
```
nschonni, line 17: I think instead of space, it needs to be all characters, except the ones you've already identified in the next line. That captures stuff like the emoji characters in the titles that are failing in some of the tests right now

DavidAnson:
Agree, there is kind of a lot to this! :) But the issues it already found make the value of this rule clear and I plan to continue working on this to include it with markdownlint. You've got this off to a great start and it's in a pretty good place right now. If you're okay with it, what I'm thinking is to merge your work into a dedicated branch in the official repo and I'll take over. I want to add a/name support probably and I want to customize the TestRepos test to temporarily check with another tool and I want to draft a set of issues for some of the test repos to get them looking into things soon. I'm happy to do all that and I think it may be more efficient for me to do than you and I trying to go back and forth here in the PR comments. :) Please let me know if that works for you.

PS - I'll be sure the commit with your work stays in your name so you'll get credit on GitHub for the contribution.

lib/md051.js
nschonni: Maybe it is `\W` https://www.tutorialspoint.com/w-vs-w-in-javascript-regex

DavidAnson: The particular tests you're worried about seem to be passing already? Unless you mean that GitHub translates the links differently and the tests aren't catching that?

nschonni: I mean it seems the repo tests for Mocha are failing from the emoji characters in https://github.com/mochajs/mocha/blob/master/.github/CONTRIBUTING.md
I'm not sure if there is a difference between GFM and CommomMark for those characters

nschonni: Actually, I might have it backwards. Clicking on the link on github gets https://github.com/mochajs/mocha/blob/master/.github/CONTRIBUTING.md#bug-i-found-a-bug

Divlo: > The particular tests you're worried about seem to be passing already?

Yep, it should already work with Emojis, and there are already passing tests for that. :smile:

Divlo:
Sounds good to me! :+1: @DavidAnson 

> I'm happy to do all that and I think it may be more efficient for me to do than you and I trying to go back and forth here in the PR comments.

Indeed, that would probably be more efficient for you as you know better the codebase, and maybe more experience for doing that kind of thing. :+1:

---

> I want to add a/name support

Yes, don't forget to also add support for `id` in any HTML element. For example: 
```html
<p id='random-id'>Text</p>
``` 

It allows to have a link fragment like that: 

```md
[Valid](#random-id)
```

---

> Please let me know if that works for you.

Sure, that works! I would be happy to follow the progress for this.

DavidAnson:
FYI: I won't get to this for a little while still, but I've captured the work above in a branch of the repo just in case the fork goes away.

Divlo:
> FYI: I won't get to this for a little while still, but I've captured the work above in a branch of the repo just in case the fork goes away.

Right, I'll close the PR and delete my fork then. :+1:

DavidAnson:
@Divlo FYI, this is merged to `next` branch.

PR #496, opened 2022-01-29
Remove need for process.platform in applyFixes

Billiam:
Related to fd24b9552be8ca6eb543cf3f74fdccb44c9959cb

Also pass platform argument for `applyFixes` call to remove dependency on `process.platform` if line ending can't be determined from the document content.


DavidAnson (changes requested):
Makes sense! Please add a test and rebase onto `next` branch.

PR #539, opened 2022-07-26
Update MD051/link-fragments to support `id` attributes on non-`a` elements.

### LINKED ISSUE #538: MD051 should support arbitrary `id` attributes, not just `<a>` tags
MD051 currently only checks `<a>` tags for `id="..."` attributes, but this requirement seems arbitrary. Any HTML element may have an `id` attribute to define an anchor.

It seems like this can be easily fixed by removing the check for `a` tags on [line 58](https://github.com/DavidAnson/markdownlint/blob/cba2ca0dbdc8f11e4b97c054380ffd52bb7aae19/lib/md051.js#L58).

In my case, I have a table and use `id` attributes to link to specfic rows of the table:

```markdown
# Example

<table>
<tr id="foo"><td>Foo<td>Foo is good
<tr id="bar"><td>Bar<td>Bar is also good
</table>

I like [foo](#foo) and [bar](#bar).
```

```
$ npx markdownlint test.md
test.md:8:8 MD051/link-fragments Link fragments should be valid [Context: "[foo](#foo) and [bar](#bar)"]
test.md:8:8 MD051/link-fragments Link fragments should be valid [Context: "[foo](#foo)"]
```

Would you be willing to accept a PR that fixes this, or is there some other reason why only `a` tags are accepted?

MarkLodato:
The `name` is only an anchor on `a` elements, but `id` is a universal
attribute on all elements.

Fixes #538.


demo/markdownlint-browser.js
```diff
@@ -4643,11 +4644,10 @@ module.exports = {
             let match = null;
             while ((match = htmlElementRe.exec(token.content)) !== null) {
                 const [tag, , element] = match;
-                if (element.toLowerCase() === "a") {
-                    const idMatch = identifierRe.exec(tag);
-                    if (idMatch) {
-                        fragments.set(`#${idMatch[1]}`, 0);
-                    }
+                const idMatch = identifierRe.exec(tag) ||
```
nschonni, line 4647: maybe `anchorMatch` would be better since it's not just IDs

DavidAnson: It's always matching on "id", just not limiting to "a" tags.

lib/md051.js
```diff
@@ -6,7 +6,8 @@ const { addError, escapeForRegExp, filterTokens, forEachInlineChild,
   forEachHeading, htmlElementRe } = require("../helpers");
 
 // Regular expression for identifying HTML anchor names
-const identifierRe = /(?:id|name)\s*=\s*['"]?([^'"\s>]+)/iu;
```
DavidAnson, line 9: What about leaving it as a single RegExp, making id/name a capture, and looking at that?

demo/markdownlint-browser.js
DavidAnson: I think I get your point now, open to this.

test/link-fragments.md
```diff
@@ -125,6 +127,10 @@ Text
 
 <A href="https://example.com" id="HREFandID">Text</A>
 
+<p id="paragraph-id"></p>
+
+<p name="not-an-anchor"></p>
```
DavidAnson, line 132: Please add another case that has p/name to confirm that is NOT allowed.

DavidAnson (changes requested):
Thanks! Please review comments and send PR against "next" branch.

demo/markdownlint-browser.js
nschonni: maybe `anchorMatched` is a little clearer

lib/md051.js
MarkLodato: That would only capture one or the other, but it is valid to have both. For example, with `<input name=foo id=bar>`, the anchor should be `bar`, not `foo`. I could make it a global regexp, but I that would complicate the code (having to loop). I thought the two regexps, though more verbose, was a bit easier to follow.

But if you disagree, let me know and I'll give it a shot.

(I just added a test case to cover such a potential regression.)

demo/markdownlint-browser.js
MarkLodato: Agreed. Done. Went with `anchorMatch` since this is a "match" object, not a "matched" boolean.

test/link-fragments.md
MarkLodato: Sorry, I don't understand. This case is checked on line 148 to confirm that the `name` tag is not accepted on a non-`a` element. Had you missed that, or are you asking for a different case?

By the way, I changed the names to make it more obvious what is being tested.

MarkLodato (commented):
> send PR against "next" branch.

Oops! Fixed.

I did not rebase so that it's easier to review on GitHub. (If you rebase, GitHub completely loses track of comments.) I'll rebase/squash once all the comments are resolved.

lib/md051.js
DavidAnson: I hadn't thought of both being present. What you've done is good!

lib/md051.js
```diff
@@ -6,7 +6,8 @@ const { addError, escapeForRegExp, filterTokens, forEachInlineChild,
...
+const identifierRe = /id\s*=\s*['"]?([^'"\s>]+)/iu;
+const nameRe = /name\s*=\s*['"]?([^'"\s>]+)/iu;
```
DavidAnson, line 10: Should we put \w at the beginning to block "data-id" from matching?

demo/markdownlint-browser.js
```diff
@@ -4601,7 +4601,8 @@ module.exports = [
 
 const { addError, escapeForRegExp, filterTokens, forEachInlineChild, forEachHeading, htmlElementRe } = __webpack_require__(/*! ../helpers */ "../helpers/helpers.js");
 // Regular expression for identifying HTML anchor names
-const identifierRe = /(?:id|name)\s*=\s*['"]?([^'"\s>]+)/iu;
+const identifierRe = /id\s*=\s*['"]?([^'"\s>]+)/iu;
```
DavidAnson, line 4604: Can we call this "idRe" so it's consistent?

DavidAnson (changes requested):
Cool, getting close!

demo/markdownlint-browser.js
MarkLodato: Done.

lib/md051.js
MarkLodato: I noticed that too. Done and added a unit test.

(I think you meant `\s`. Note that `\b` doesn't work because "-" is not a word character.)

DavidAnson: Right, I meant word boundary, but good call on '-'.

DavidAnson:
Looks great! I will do a final review from my desktop machine this evening and merge the PR. Thank you for your attention to detail and promptness!

MarkLodato:
Thanks! Rebased - should be ready to merge.

DavidAnson (approved):

PR #587, opened 2022-09-23
doc: clarify description for MD026 punctuation

nschonni:
Was trying to customize the rule on a repo and couldn't easily tell from the current metadata if it was an allow or a disallow list of characters.

There is a similar ambiguity in the MD036 rule

### DIFF
--- a/schema/.markdownlint.jsonc
+++ b/schema/.markdownlint.jsonc
@@ -139,7 +139,7 @@
 
   // MD026/no-trailing-punctuation - Trailing punctuation in heading
   "MD026": {
-    // Punctuation characters
+    // Punctuation characters not allowed at end of headings
     "punctuation": ".,;:!。，；：！"
   },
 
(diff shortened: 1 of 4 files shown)


schema/.markdownlint.jsonc
```diff
@@ -139,7 +139,7 @@
 
   // MD026/no-trailing-punctuation - Trailing punctuation in heading
   "MD026": {
-    // Punctuation characters
+    // Punctuation characters to remove
```
DavidAnson, line 142: Thanks! To be picky, it's only removed when fixing. What about using the word "block" or "prevent" instead?

DavidAnson: Also, any reason not to do the same improvement for the other rule?

DavidAnson (changes requested):

schema/.markdownlint.jsonc
nschonni: I was less clear on the use in MD036 when I was reading the code, so I wasn't sure about the wording.
I'll take another cut at the wording on this one though

DavidAnson (approved):
Cool

Issue #593, opened 2022-10-02
Docs: document how to extend from a built-in style


JoshuaKGoldberg:
Following #592: https://github.com/DavidAnson/markdownlint#optionsconfig mentions that you can extend from premade styles and includes an example of how to do so with a CommonJS `require()`. But, the docs don't show:

* How to do so in a JSON file
* How to extend from one of the prebuilt styles in https://github.com/DavidAnson/markdownlint/blob/main/style

For example, I tried this in a JSON config file:

```json
{
  "extends": "markdownlint/style/all.json"
}
```

...but that didn't do anything.

Could this be documented, please? I'm happy to send in docs myself... if someone tells me how to do these things 😄

DavidAnson:
In general, an explicit path or a module identifier can be specified. In this case, the somewhat recent addition of `package.json/exports` may block the syntax you show. Module resolution will also depend on the context in which it is invoked. The most reliable mechanism, may be to copy one of the example styles into your project and reference it from there.

Issue #579, opened 2022-09-11
MD033 flagging HTML tags in image alt text strings


nschonni:
https://dlaa.me/markdownlint/#%25m!%5BThe%20default%2C%20focused%2C%20and%20disabled%20%3Ctextarea%3E%20element%20in%20Firefox%2071%20and%20Safari%2013%20on%20Mac%20OSX%20and%20Edge%2018%2C%20Yandex%2014%2C%20Firefox%20and%20Chrome%20on%20Windows%2010.%5D(textarea_basic.png)
Since the alt tag doesn't get parsed as HTML, there shouldn't be a need to escape these.
Ran into this because i had been escaping the tags, but then running prettier would clean off the escaping because it was not needed.

Similar thing happens with link title strings, but that's probably a separate bug

DavidAnson:
As your example shows, HTML content in the image alternate text region can be removed by the parser and so I think it is reasonable for markdownlint to warn about it.

Here is an example using markdown-it directly: http://markdown-it.github.io/#md3=%7B%22source%22%3A%22%23%20Issue%20579%5Cn%5Cn!%5Btext%20%3Ctextarea%3E%20text%5D%28image.png%29%5Cn%22%2C%22defaults%22%3A%7B%22html%22%3Atrue%2C%22xhtmlOut%22%3Afalse%2C%22breaks%22%3Afalse%2C%22langPrefix%22%3A%22language-%22%2C%22linkify%22%3Atrue%2C%22typographer%22%3Atrue%2C%22_highlight%22%3Atrue%2C%22_strict%22%3Afalse%2C%22_view%22%3A%22src%22%7D%7D

nschonni:
Hmm, I'm thinking it might be a Markdown-it bug then. If you run it through GitHub's parser or the remark parser like Prettier uses, its not treated as a literal

![The default, focused, and disabled <textarea> element in Firefox 71 and Safari 13 on Mac OSX and Edge 18, Yandex 14, Firefox and Chrome on Windows 10.](textarea_basic.png)

DavidAnson:
Toggling the "HTML" checkbox on that demo page opt into and out of this removal behavior.

Skimming the CommonMark specification, it's not clear to me that this scenario is directly addressed, so I think the parser is behaving consistently.

nschonni:
I filed something on Markdown-it, but looking at the spec https://spec.commonmark.org/0.30/#images it is light, but 

[...]

PR #594, opened 2022-10-02
feat: added style/prettier.json

### LINKED ISSUE #592: Add an option to disable all Prettier-related rules at once
I often use Markdownlint and Prettier together in the same repo, both with editor extensions. When using Prettier, there's no need for Markdownlint to check formatting rules. I find myself disabling at least `no-hard-tabs` and `line-length`, and sometimes more if the Prettier config is customized (docs ref: https://github.com/DavidAnson/markdownlint/blob/main/doc/Prettier.md).

Can we have a one-line option in for Markdownlint configs that disables all formatting rules that would conflict with Prettier? For reference, the equivalent in ESLint-land is [`eslint-config-prettier`](https://github.com/prettier/eslint-config-prettier).

* Example repo: https://github.com/JoshuaKGoldberg/template-typescript-node-package/tree/19d02d06310097e867344714b36300776f12721a.
* Markdownlint configuration: https://github.com/JoshuaKGoldberg/template-typescript-node-package/blob/19d02d06310097e867344714b36300776f12721a/.markdownlint.json

If the answer to the issue is to disable all rules with certain tags, then this might be a docs request, not a feature request. 😄

JoshuaKGoldberg:
Fixes #592.

I wasn't sure from the phrasing whether the issue is accepting PRs, but figured this was a good way to demonstrate the list of what would need to be done in a `prettier.json`.


style/prettier.json
```diff
@@ -0,0 +1,25 @@
+{
+  "comment": "Disables rules that may conflict with Prettier",
+  "blanks-around-fences": false,
```
nschonni, line 3: These rules aren't in conflict, as Prettier adds these too.
It passes the code through https://prettier.io/docs/en/options.html#parser `remark-parse`
Most of the rules below aren't configurable with prettier, but don't conflict with Markdownlint

nschonni (commented):
Most of these aren't in conflict between Prettier and Markdownlint.
Implemented both in a large repo https://github.com/mdn/content there have been few conflicts outside some underspec'd CommonMark things like https://github.com/DavidAnson/markdownlint/issues/579 that are implemented differently in markdown-it (Markdownlint dep) and remark-parse (prettier dep)

style/prettier.json
JoshuaKGoldberg: What is the point in having these rules enabled in Markdownlint-land, though? If you have editor extensions enabled for both Markdownlint and Prettier, you'll get reports from both tools about the errors. That seems excessive.

DavidAnson (changes requested):
Thanks!

Info about contributions: https://github.com/DavidAnson/markdownlint/blob/main/CONTRIBUTING.md

In particular, please PR for `next` branch.

Also, please update the existing tests that sanity check the content of the `styles` directory.

style/prettier.json
DavidAnson: I don't use the Prettier tool, so I won't have many opinions about which rules should be listed here. However, I wonder if there is a distinction to be made between what rules are used by the editor plug-in, and what rules are applied on the command-line. In my practice, the CI process is definitive and people can choose to use editor plug-ins, or not. I wouldn't want to set up the rules here such that everyone was required to use both plug-ins in order to write acceptable Markdown.

DavidAnson (approved):
This looks good to me, thank you! Before completing the PR, I would like to hear again from Nick and you on the inline discussion about which rules should be excluded. If everyone is happy these are the right rules, that works for me.

style/prettier.json
JoshuaKGoldberg: > I wouldn't want to set up the rules here such that everyone was required to use both plug-ins in order to write acceptable Markdown.

Me neither 😄 - I don't think that's what this change would cause. All that's happening is folks who _opt in_ to extending `style/prettier.json` won't have Markdownlint apply them.

A more accurate phrasing from my comment in this thread might have been: If you have _both Markdownlint and Prettier configured_, you'll get reports from both tools about the errors. That seems excessive.

Am I missing something?

DavidAnson: Makes sense to me! I'll give a day or two for Nick to weigh in if he wants and then merge the PR.

nschonni: It isn't something I'd use/recommend on a project, but I don't think that should block this landing.
Probably need to update the existing documentation though https://github.com/DavidAnson/markdownlint/blob/a852407c887ec60949aa5365ed964bab833f962f/doc/Prettier.md

DavidAnson:
FYI, as part of fixing #593, you'll need to do a trivial merge and edit to `package.json` for this PR.

DavidAnson:
I applied the merge conflict in the web editor just now.

DavidAnson (changes requested):
Un-approving until I review the new changes to Prettier.md.

doc/Prettier.md
```diff
@@ -1,8 +1,11 @@
 # Using markdownlint with Prettier
 
-[`Prettier`](https://prettier.io) is a popular code formatter. For the most part,
-it works seamlessly with `markdownlint`. Special situations are documented
-below.
+[`Prettier`](https://prettier.io) is a popular code formatter.
+For the most part, Prettier works seamlessly with `markdownlint`.
+
+You can extend from the `markdownlint/style/prettier.json` style to disable all Markdownlint rules that overlap with Prettier.
```
DavidAnson, line 6: Please:
- wrap at 80 columns
- lower-case "markdownlint" (breaking CI)
- remove word "from"
- add "rules" to the end: "overlap with Prettier rules"

DavidAnson (changes requested):

doc/Prettier.md
JoshuaKGoldberg: Nit: on the last point, my understanding is Prettier doesn't really have "rules" per se. Unless I'm wrong it's a one-pass AST printing that just has a few configuration settings.

DavidAnson (approved):
Nice, thanks!

DavidAnson:
Closed by https://github.com/DavidAnson/markdownlint/commit/82a5de08212ce5004d3bf2a923786ec4cb19b73b (PR required manual merge). Thanks for the contribution!

Issue #608, opened 2022-10-16
Support `allowlist` option for MD040


chenxsan:
While MD040 requires us to specify a language, it doesn't specify which languages we can use, which means we can use whatever we want as long as it's not an empty one.

But in the cases of syntax highlighting, tools like [refractor](https://github.com/wooorm/refractor#data) will throw an error when we feed it an unknown language. It would be great if we can configure MD040 with an `allowlist`.

E.g.,

```
"MD040": { "allowlist": ["js", "css", "html"] },
```

nschonni:
Duplicate of https://github.com/DavidAnson/markdownlint/issues/608

PR #609, opened 2022-10-16
Add allowlist parameter to MD040

chenxsan:
fixes https://github.com/DavidAnson/markdownlint/issues/608.


chenxsan:
This one was based on `main` branch not `next`, hence closed.

PR #610, opened 2022-10-16
Add allowed_languages parameter to MD040

chenxsan:
fixes https://github.com/DavidAnson/markdownlint/issues/608.


chenxsan:
I just kept running into this error when running `npm test`:

```
markdownlint-scenarios › md040-allowlist.md

  test/markdownlint-test-scenarios.js:68

   67:         }                                                                  
   68:         t.deepEqual(actual, expected, "Too few or too many issues found.");
   69:         // Create snapshot                                                 

  Too few or too many issues found.

  Difference:

    {
  -   MD040: [
  -     5,
  -     11,
  -   ],
    }

  › test/markdownlint-test-scenarios.js:68:11

  ─

  1 test failed
```

@DavidAnson Any idea how can I fix it?

chenxsan:
`npm run ci` would show me:

```
> markdownlint@0.26.2 build-demo
> node scripts copy node_modules/markdown-it/dist/markdown-it.min.js demo/markdown-it.min.js && cd demo && webpack --no-stats

diff --git a/demo/markdownlint-browser.js b/demo/markdownlint-browser.js
index 8ff0112..c0d19a6 100644
--- a/demo/markdownlint-browser.js
+++ b/demo/markdownlint-browser.js
@@ -4152,15 +4152,22 @@ module.exports = {
 "use strict";
 // @ts-check
 
-const { addErrorContext, filterTokens } = __webpack_require__(/*! ../helpers */ "../helpers/helpers.js");
+const { addErrorContext, filterTokens, addError } = __webpack_require__(/*! ../helpers */ "../helpers/helpers.js");
 module.exports = {
     "names": ["MD040", "fenced-code-language"],
     "description": "Fenced code blocks should have a language specified",
     "tags": ["code", "language"],
     "function": function MD040(params, onError) {
+        let allowList = params.config.allow_list;
+        allowList = Array.isArray(allowList) ? allowList : null;
         filterTokens(params, "fence", function forToken(token) {
-            if (!token.info.trim()) {
+            const lang = token.info.trim();
+            if (lang === "") {
                 addErrorContext(onError, token.lineNumber, token.line);
+                return;
+            }
+            if (allowList !== null && allowList.includes(lang) === false) {
+                addError(onError, token.lineNumber, `language "${lang}" is not included in allow_list`);
             }
         });
     }
```
Not sure how should I fix it as there's no such infomations in https://github.com/DavidAnson/markdownlint/blob/main/CONTRIBUTING.md

lib/md040.js
```diff
@@ -2,16 +2,27 @@
...
-const { addErrorContext, filterTokens } = require("../helpers");
+const { addErrorContext, filterTokens, addError } = require("../helpers");
 
 module.exports = {
   "names": [ "MD040", "fenced-code-language" ],
   "description": "Fenced code blocks should have a language specified",
   "tags": [ "code", "language" ],
   "function": function MD040(params, onError) {
+    let allowList = params.config.allow_list;
+    allowList = Array.isArray(allowList) ? allowList : null;
```
nschonni, line 13: I think other rules us an empty array as a default https://github.com/DavidAnson/markdownlint/pull/610/files#diff-b9c3f0adb939e1b4f7e9c2e963f2f49f6f09a4a5fd71fd4a7a4811732d69aff0R408-R412
Think it might make the code parsing simpler

DavidAnson:
Please include the changes to the demo file with your commit. I made myself a note to update the instructions.

doc/Rules.md
```diff
@@ -1571,6 +1571,8 @@ Tags: code, language
 
 Aliases: fenced-code-language
 
+Parameters: allow_list (array of string; default `null`)
```
DavidAnson, line 1574: What about calling it "valid_languages"? "Allow" doesn't seem as appropriate here because this rule is requiring one. "Languages" instead of "list" makes it clearer what is expected.

doc/Rules.md
```diff
@@ -1598,6 +1600,8 @@ Plain text in a code block
 ```
 ````
 
+You can configure the `allow_list` parameter to specify a list of languages your code blocks could use. The default value is `null` which means any language specifier is all right.
```
DavidAnson, line 1603: Remove "your", change "all right" to "valid", maybe?

lib/md040.js
```diff
@@ -2,16 +2,27 @@
 
 "use strict";
 
-const { addErrorContext, filterTokens } = require("../helpers");
+const { addErrorContext, filterTokens, addError } = require("../helpers");
```
DavidAnson, line 5: Sort, please.

lib/md040.js
DavidAnson: Agree, see MD033.

lib/md040.js
```diff
@@ -2,16 +2,27 @@
...
   "tags": [ "code", "language" ],
   "function": function MD040(params, onError) {
+    let allowList = params.config.allow_list;
+    allowList = Array.isArray(allowList) ? allowList : null;
     filterTokens(params, "fence", function forToken(token) {
-      if (!token.info.trim()) {
+      const lang = token.info.trim();
+      if (lang === "") {
         addErrorContext(onError, token.lineNumber, token.line);
+        return;
```
DavidAnson, line 18: Replace with else.

lib/md040.js
```diff
@@ -2,16 +2,27 @@
...
+      }
+      if (allowList !== null && allowList.includes(lang) === false) {
```
DavidAnson, line 20: Parenthesis for each term, don't compare against Boolean, please.

lib/md040.js
```diff
@@ -2,16 +2,27 @@
...
+        addError(
+          onError,
+          token.lineNumber,
+          `language "${lang}" is not included in allow_list`
```
DavidAnson, line 24: Simplify to "$lang is not valid"

test/md040-allowlist.md
```diff
@@ -0,0 +1,29 @@
+# fenced_code_blocks
```
DavidAnson, line 1: Love the tests! My new convention is to make the top level heading correspond to the file name.

DavidAnson (changes requested):
Looks very thorough, thank you! I left a few comments here and may have a couple more once I review from a computer. Appreciate the contribution!

DavidAnson:
I changed my mind. I think "allowed_languages" is the right parameter name. So the error message should say "$language is not allowed".

lib/md040.js
chenxsan: I used empty array as a default in my first implementation, but I thought it would mean no language was allowed semantically, that's why I changed it to `null` later. What do you think?

nschonni: https://github.com/DavidAnson/markdownlint/blob/2b2dc27f24a16a2afdf66d850a247f7ce02df6c8/lib/md033.js#L21-L23

lib/md040.js
chenxsan: Eslint won't let me use `else`:




Should I disable the rule?

DavidAnson: `if (lang) {...} else if (allowed && includes) {...}` should be allowed.

doc/Rules.md
```diff
@@ -1598,6 +1600,8 @@ Plain text in a code block
 ```
 ````
 
+You can configure the `allowed_languages` parameter to specify a list of languages code blocks could use. The default value is `null` which means any language specifier is valid.
```
DavidAnson, line 1603: I think it will want you to wrap this line at 80 characters?

lib/md040.js
```diff
@@ -2,16 +2,30 @@
...
 
 module.exports = {
   "names": [ "MD040", "fenced-code-language" ],
   "description": "Fenced code blocks should have a language specified",
   "tags": [ "code", "language" ],
   "function": function MD040(params, onError) {
+    let allowed = params.config.allowed_languages;
+    allowed = Array.isArray(allowed) ?
+      allowed.map((lang) => lang.toLowerCase()) :
+      null;
```
DavidAnson, line 15: Like Nick suggested, using [] instead of null avoids the need to compare against null on line 22.

test/md040-allowed_languages.json
```diff
@@ -0,0 +1,5 @@
+{
+  "default": true,
```
DavidAnson, line 2: This is unnecessary and I'd like to stop using it - please remove?

lib/md040.js
```diff
@@ -2,16 +2,30 @@
...
+const { addError, addErrorContext, filterTokens } = require("../helpers");
 
 module.exports = {
   "names": [ "MD040", "fenced-code-language" ],
   "description": "Fenced code blocks should have a language specified",
   "tags": [ "code", "language" ],
   "function": function MD040(params, onError) {
+    let allowed = params.config.allowed_languages;
+    allowed = Array.isArray(allowed) ?
+      allowed.map((lang) => lang.toLowerCase()) :
```
DavidAnson, line 14: Thanks for adding toLowerCase!

test/md040-allowed_languages.md
```diff
@@ -0,0 +1,45 @@
...
+
+Code block with `md` in allowed_languages:
+
+```MD
+hello md
+```
+
+Code block with `TS` in allowed_languages:
+
+```ts
```
DavidAnson, line 39: Nice having one test case each for upper/lower case.

DavidAnson (changes requested):
Very minor feedback to follow-up here, thanks!

doc/Rules.md
DavidAnson: Looks like the tests pass, but it would still be good to do this to be consistent with other text in the file.

test/md040-allowed_languages.json
DavidAnson: To clarify: `default:true` is not needed.

DavidAnson:
> Please include the changes to the demo file with your commit. I made myself a note to update the instructions.

I went to do this and It's there already, line 39, "As part of a continuous integration run, generated files may get updated and fail the run - commit them to the repository and rerun continuous integration.".

chenxsan:
> > Please include the changes to the demo file with your commit. I made myself a note to update the instructions.
> 
> I went to do this and It's there already, line 39, "As part of a continuous integration run, generated files may get updated and fail the run - commit them to the repository and rerun continuous integration.".

I've run `npm run build-demo` to fix it.

DavidAnson (approved):
Nice!

Issue #613, opened 2022-10-19
MD043 option for match heading case


mateus:
Currently, all headings passed to the [MD043](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md043) rule  are not case sensitive. Meaning that `This Heading` and `this heading` are both seen as valid for this rule. I would like to propose the addition of an option that will consider the case of the heading as well. Usage would look like this:

```
MD043: {
  headings: ['*', '## Lorem ipsum dolor', '## Consectetur adipiscing elit'],
  match_case: true,
},
```

The `match_case` option can be set to false as the default value to prevent breaking any existing usage of this rule.

PR #614, opened 2022-10-19
Add match_case to MD043

mateus:
Fixes https://github.com/DavidAnson/markdownlint/issues/613


test/required-headings-match-case.json
```diff
@@ -0,0 +1,13 @@
...
+  "default": true,
+  "MD043": {
+    "headings": [
+      "# Title",
+      "## First Heading",
+      "## Second Heading",
+      "*",
+      "## Third Heading"
+    ],
+    "match_case": true
```
DavidAnson, line 11: Thanks for adding tests! Please add something to prove this rule will pass on mixed case headings by default and fail on similar input when the new parameter is set.

lib/md043.js
```diff
@@ -24,17 +25,23 @@ module.exports = {
       forEachHeading(params, (heading, content) => {
         if (!hasError) {
           anyHeadings = true;
-          const actual = levels[heading.tag] + " " + content;
-          const expected = getExpected();
+          const actual = matchCase ?
+            levels[heading.tag] + " " + content :
+            levels[heading.tag] + " " + content.toLowerCase();
```
DavidAnson, line 30: I'd like to avoid the duplication that's been added. What about adding a helper like `const handleCase = (s) => match_case ? s : s.toLowerCase()`? Then wrap `content`, etc. with that.

DavidAnson (changes requested):
Thanks for the contribution! Two requests from initial review.

test/required-headings-match-case.json
mateus: I initially thought that [this test](https://github.com/DavidAnson/markdownlint/blob/main/test/markdownlint-test.js#L631-L687) was sufficient for the cases where the option is turned off (or it is simply not specified). 

Can you share an example of how I can achieve what you mean, please? I am not very familiar with the structure of these tests. Are you suggesting to add a new unit test asserting that the rule fails when the input does not match?

DavidAnson: The file you added has case in the content match the way the rule is configured and match_case is true. Please add another file where the content does not match case and match_case is true. There should be failures reported which you can tag via the "{MD043}" syntax to show they are expected. There should be a third file (it should already exist, but please add it if not) to verify that content which does not match case is allowed when match_case is not set (i.e., to verify that the code before you changed it allows case to match or not match).

mateus: I added a new file, can you double check the name and content, please? For the third file, I wonder if https://github.com/DavidAnson/markdownlint/blob/main/test/required-headings-all-present.md is covering that already. What do you think?

DavidAnson: Agreed, that file covers the default behavior of allowing any case.

test/required-headings-match-case.json
```diff
@@ -0,0 +1,13 @@
+{
+  "default": true,
```
DavidAnson, line 2: This is unnecessary here and in the other JSON file.

test/required-headings-wrong-match-case.md
```diff
@@ -0,0 +1,9 @@
+# Title
+
+## First heading {MD043}
```
DavidAnson, line 3: Could you please do all lower for this?

test/required-headings-wrong-match-case.md
```diff
@@ -0,0 +1,9 @@
...
+
+## Second Heading
+
+### Random heading
+
+## Third Heading
```
DavidAnson, line 9: And all upper for this (adding "{MD043}"?

test/snapshots/markdownlint-test-scenarios.js.md
```diff
@@ -5741,7 +5741,7 @@ Generated by [AVA](https://avajs.dev).
         },
         {
           errorContext: null,
-          errorDetail: 'Expected: # Broken; Actual: # Heading 3 {MD003} {MD043}',
+          errorDetail: 'Expected: # broken; Actual: # heading 3 {md003} {md043}',
```
DavidAnson, line 5744: Aha, this looks like a bug. The message here should use the original text, not the text after applying toLowerCase().

lib/md043.js
```diff
@@ -21,20 +22,21 @@ module.exports = {
       let hasError = false;
       let anyHeadings = false;
       const getExpected = () => requiredHeadings[i++] || "[None]";
+      const handleCase = (str) => (matchCase ? str : str.toLowerCase());
       forEachHeading(params, (heading, content) => {
         if (!hasError) {
           anyHeadings = true;
-          const actual = levels[heading.tag] + " " + content;
-          const expected = getExpected();
+          const actual = levels[heading.tag] + " " + handleCase(content);
```
DavidAnson, line 29: Based on the bug I call out in the output, it may be necessary to leave actual and expected alone and apply handleCase() in the comparisons below so actual and expected have the original case when passed to addErrorDetailIf.

DavidAnson (changes requested):
Thank you! The new file you added is good and the test I asked about was already present. I think we've spotted a bug in the output and I left a comment about that. To resolve the merge conflict with the snapshot, you should be able to rebase onto the latest `next` branch and then `npm run update-snapshots` during the merge conflict and `git rebase --continue`. Do another update-snapshots when done to confirm.

test/required-headings-wrong-match-case.md
DavidAnson: I pushed a commit for this - I'd forgotten only the first wrong heading gets flagged.

DavidAnson (approved):
Thanks again!

mateus:
@DavidAnson thoughts on when this might be released? Thanks!

BrentHuang:
how to fix the default MD043 rule?

DavidAnson:
@BrentHuang Not sure what your comment here is, but there are examples of using MD043 in the repo, such as: https://github.com/DavidAnson/markdownlint/blob/main/test/required-headings-all-present.md?plain=1

PR #628, opened 2022-11-05
fix: Make MD040/fenced-code-language case sensitive.

### LINKED ISSUE #627: Add an option allowed_languages to the fenced-code-language
Add an option _allowed_languages_ to the [_fenced-code-language_](https://github.com/DavidAnson/markdownlint/blob/a6489ac/doc/Rules.md#md040) rule to define the allowed languages. With this option, `markdownlint` will be able to check:

- name consistency (`javascript` / `js`);
- case (`html` / `HTML`);
- typos (`markdown` / `markdonw`)
- unsupported languages (`json` / `jfather`).

regseb:
Fix #627.


lib/md040.js
```diff
@@ -10,17 +10,12 @@ module.exports = {
...
+    allowed = Array.isArray(allowed) ? allowed : [];
     filterTokens(params, "fence", function forToken(token) {
       const lang = token.info.trim();
       if (lang === "") {
         addErrorContext(onError, token.lineNumber, token.line);
-      } else if (
-        allowed.length > 0 &&
-        !allowed.includes(lang.toLowerCase())
-      ) {
+      } else if (allowed.length > 0 && !allowed.includes(lang)) {
```
DavidAnson, line 18: Please add parents around the length comparison.

test/md040-allowed_languages.md
```diff
@@ -12,6 +12,18 @@ Code block with `css` not in allowed_languages:
 body {}  {MD040:11}
 ```
 
+Code block with `MD` (uppercase) not in allowed_languages:
```
DavidAnson, line 15: Nice, thanks for adding both to be sure!

DavidAnson (changes requested):
This is great, thank you for doing this!

Looks like the generated file needs to be added to pass CI. Could you also please make a tiny update to doc-build/md040.md to mention case is sensitive for this rule?

regseb:
I added the parenthesis and a sentence in the documentation. I ran all `build-*`.

I also removed the trim for comparison with allowed languages so this code reports an error:

````markdown
<!-- markdownlint-configure-file { "MD040": { "allowed_languages": ["js"] } } -->
``` js
// ^ Space!
let foo = "bar";
```
````

DavidAnson:
The specifications says that the info string is trimmed of leading and trailing whitespace, so I do not think it is appropriate to remove the call to `trim`: https://spec.commonmark.org/0.30/#fenced-code-blocks

nschonni:
Might need a test for extra metadata like https://spec.commonmark.org/0.30/#example-143 or just validate that the first token is in the allowlist, and ignore the rest

DavidAnson:
Great point! I am happy to add that myself in a separate commit, or it would be fine to include here.

regseb:
I re-added the trim and only check the first word.

I would like to have a way to compare all the _info string_ with the allowed languages. Maybe add a `strict` option that bypass `.trim().split(/\s+/u).shift()`?

DavidAnson:
Only the first "word" of the info string is the language, right? That should be present so the code can be formatted properly. But the rest of the line is a mystery. It may be used to format better, but I'd worry about a setting that blocks if it may be needed for some cases?

DavidAnson (approved):
Thanks for the contribution!

PR #629, opened 2022-11-06
feat(fenced-code-language): Add option to refuse extra data

regseb:
Add option to refuse leading and trailing whitespace; and extra data in fence code blocks information.

PR related:  #628 and #610


demo/markdownlint-browser.js
```diff
@@ -4176,6 +4178,9 @@ module.exports = {
             else if ((allowed.length > 0) && !allowed.includes(lang)) {
                 addError(onError, token.lineNumber, `"${lang}" is not allowed`);
             }
+            if (!extra && token.info !== lang) {
```
DavidAnson, line 4181: Parenthesis, please.

demo/markdownlint-browser.js
```diff
@@ -4168,6 +4168,8 @@ module.exports = {
     "function": function MD040(params, onError) {
         let allowed = params.config.allowed_languages;
         allowed = Array.isArray(allowed) ? allowed : [];
+        let extra = params.config.extra;
+        extra = (extra === undefined) ? true : !!extra;
```
DavidAnson, line 4172: I prefer to default unset parameters to false. Could this be reframed as "language_only"? (Not sure I love that name, but you get the idea.)

doc-build/md040.md
```diff
@@ -29,6 +29,12 @@ You can configure the `allowed_languages` parameter to specify a list of
 languages code blocks could use. Languages are case sensitive. The default value
 is `[]` which means any language specifier is valid.
 
+You have the option to refuse extra data in fence code blocks information. To do
```
DavidAnson, line 32: I may have future comments on the phrasing here, but will wait until we agree on the parameter name.

test/md040-extra.md
```diff
@@ -0,0 +1,25 @@
+# md040-extra.md
+
+Fence code block information with leading whitespaces:
+
+```  html
+<h1>markdownlint</h1>  {MD040:5}
```
DavidAnson, line 6: Nit: Extra space

test/md040-extra.md
```diff
@@ -0,0 +1,25 @@
+# md040-extra.md
```
DavidAnson, line 1: Love the new test cases!

test/md040-extra.md
```diff
@@ -0,0 +1,25 @@
...
+Fence code block information with leading whitespaces:
+
+```  html
+<h1>markdownlint</h1>  {MD040:5}
+```
+
+Fence code block information with trailing whitespace:
+
+```css 
+body {}  {MD040:11} {MD009:11}
```
DavidAnson, line 12: It's not unreasonable, but should we really warn for both rules in this scenario? Per the specification, trailing white space will already be removed and so is harmless and arguably not "extra" data. I'm thinking maybe only the trailing white space rule should report for this example. What do you think?

DavidAnson (changes requested):
I think this is a good idea, and I like the behavior you have implemented! I left a few comments about how this PR might be tweaked. Thanks!

regseb:
My proposals for the name of the parameter (from best to worst): `strict`, `language_only`, `no_extra`, `no_extra_data`, `one_word`, `single`.

---

With the `language_only` option, I expect the first line of the fenced code block to contain only the language (without spaces).

The two rules `no-trailing-spaces` and `fenced-code-language` check the same thing, but they don't have the same scope. Maybe a user only wants to check the language strictly. If we don't check the trailing spaces in the _info string_: the user won't get the information about the extra spaces.

The best thing to do would be to add a parameter to the `no-trailing-spaces` rule to not check for trailing spaces for fences. Personally, I wouldn't activate this parameter because I don't mind having two errors for the same case. I think this option would be superfluous.

DavidAnson:
I don't like `strict` for this, so let's go with `language_only` - and it looks like you've already made that change. I'm fine with the overlap in trailing spaces we have now. I'll make a second review pass shortly. Thanks!

lib/md040.js
```diff
@@ -11,6 +11,8 @@ module.exports = {
   "function": function MD040(params, onError) {
     let allowed = params.config.allowed_languages;
     allowed = Array.isArray(allowed) ? allowed : [];
+    const only = !!params.config.language_only;
```
DavidAnson, line 14: Please call it `languageOnly` for consistency.

lib/md040.js
```diff
@@ -22,6 +24,14 @@ module.exports = {
           `"${lang}" is not allowed`
         );
       }
+
+      if (only && (token.info !== lang)) {
+        addError(
+          onError,
+          token.lineNumber,
+          `"${token.info}" doesn't only contain the language`
```
DavidAnson, line 32: Info string contains more than language: "${token.info}"

schema/build-config-schema.js
```diff
@@ -356,6 +356,11 @@ for (const rule of rules) {
             "type": "string"
           },
           "default": []
+        },
+        "language_only": {
+          "description": "Accept language only",
```
DavidAnson, line 361: "Accept" -> "Require"

test/md040-language_only.md
```diff
@@ -0,0 +1,25 @@
+# md040-language_only.md
+
+Fence code block information with leading whitespaces:
```
DavidAnson, line 3: "whitespace" (no 's')

doc-build/md040.md
DavidAnson: How about:

You can prevent extra data from being present in the info string of fenced code blocks. To do so, set the `language_only` parameter to `true`.
 
 Info strings with leading/trailing whitespace (ex: `  js  `) or other content (ex: `ruby startline=3`) will trigger this rule.

DavidAnson (changes requested):
Looking good, just a few more comments...

test/md040-language_only.md
```diff
@@ -0,0 +1,25 @@
...
+
+```  html
+<h1>markdownlint</h1> {MD040:5}
+```
+
+Fence code block information with trailing whitespace:
+
+```css 
+body {}  {MD040:11} {MD009:11}
```
DavidAnson, line 12: Extra space

regseb:
I fixed your feedback. For the text in English, I trust you. _(Google Translate est mon ami)_

DavidAnson (approved):
Thank you!

PR #664, opened 2022-12-03
docs: add example of Node API for auto-fixing violations

### LINKED ISSUE #663: Document Node API for auto-fixing markdownlint violations
I need to fix the markdownlint violations in a string via markdownlint's Node API. I had trouble figuring out how to do this as it doesn't seem to be mentioned in markdownlint's documentation. By looking at various repositories using markdownlint's Node API, I was able to come up with this working example:

```js
const { sync } = require('markdownlint');
const markdownlintRuleHelpers = require('markdownlint-rule-helpers');
const markdownlintConfig = require('./.markdownlint.json');

function fixMarkdownlintViolations(content) {
  const fixResults = sync({
    strings: { content },
    config: markdownlintConfig,
    resultVersion: 3,
  });
  const fixes = fixResults.content.filter((error) => error.fixInfo);
  if (fixes.length > 0) {
    return markdownlintRuleHelpers.applyFixes(content, fixes);
  }
  return content;
}
```

I guess this is undocumented because of this notice in [markdownlint-rule-helpers](https://www.npmjs.com/package/markdownlint-rule-helpers)?

> Undocumented - This package exports the internal functions as-is. The APIs were not originally meant to be public, are not officially supported, and may change from release to release. There are brief descriptive comments above each function, but no [JSDoc](https://en.m.wikipedia.org/wiki/JSDoc) annotations. That said, some of what's here will be useful to custom rule authors and may avoid duplicating code.

Would it make sense to document my example in the markdownlint-rule-helpers README (which has the disclaimer about it being subject to change), perhaps linking to it from the main markdownlint README? I opened a PR to do that in #664. If there's some better place/way to document it, let me know.

bmish:
Fixes #663.

This is one possible solution to #663.

### DIFF
--- a/README.md
+++ b/README.md
@@ -930,7 +930,8 @@ bad.md: 1: MD041/first-line-heading/first-line-h1 First line in a file should be
 
 ### Fixing
 
-Rules that can be fixed automatically include a `fixInfo` property which is outlined in the [documentation for custom rules](./doc/CustomRules.md). To apply those fixes more easily, the `applyFixes` method in [markdownlint-rule-helpers](./helpers/README.md#applying-recommended-fixes) may be used.
+Rules that can be fixed automatically include a `fixInfo` property which is outlined in the [documentation for custom rules](./doc/CustomRules.md).
+To apply those fixes more easily, the `applyFixes` method in [markdownlint-rule-helpers](./helpers/README.md#applying-recommended-fixes) may be used.
 
 ## Browser
 
--- a/README.md
+++ b/README.md
@@ -928,6 +928,10 @@ bad.md: 1: MD041/first-line-heading/first-line-h1 First line in a file should be
 
 <!-- markdownlint-restore -->
 
+### Fixing
+
+Rules that can be fixed automatically include a `fixInfo` property which is outlined in the [documentation for custom rules](./doc/CustomRules.md). To apply those fixes more easily, the `applyFixes` method in [markdownlint-rule-helpers](./helpers/README.md#applying-recommended-fixes) may be used.
+
 ## Browser
 
 `markdownlint` also works in the browser.
--- a/helpers/README.md
+++ b/helpers/README.md
@@ -17,7 +17,9 @@ change from release to release. There are brief descriptive comments above each
 function, but no [JSDoc][jsdoc] annotations. That said, some of what's here will
 be useful to custom rule authors and may avoid duplicating code.
 
-## Example
+## Examples
+
+### Using Helpers from a Custom Rule
 
 ```javascript
 const { forEachLine, getLineMetadata } = require("markdownlint-rule-helpers");
@@ -39,6 +41,18 @@ module.exports = {
     });
   }
 };
+```
+
+### Applying Recommended Fixes
+
+```js
+const { markdownlintSync } = require('markdownlint');
+const markdownlintRuleHelpers = require('markdownlint-rule-helpers');
+
+function fixMarkdownlintViolations(content) {
+  const fixResults = markdownlintSync({ strings: { content } });
+  return markdownlintRuleHelpers.applyFixes(content, fixResults.content);
+}
 ```
 
 See also: [`markdownlint` built-in rule implementations][lib].


DavidAnson:
I'll look over this soon! For now, please rebase the PR against "next" branch per CONTRIBUTING.md.

helpers/README.md
```diff
@@ -42,6 +42,29 @@ module.exports = {
 };
 ```
 
+### Fixing violations
+
+An example of how to use the Node API to auto-fix violations in a string:
+
+```js
+const { sync } = require('markdownlint');
```
DavidAnson, line 50: Please name as "markdownlintSync".

helpers/README.md
```diff
@@ -42,6 +42,29 @@ module.exports = {
...
+const markdownlintRuleHelpers = require('markdownlint-rule-helpers');
+const markdownlintConfig = require('./.markdownlint.json');
+
+function fixMarkdownlintViolations(content) {
+  const fixResults = sync({
+    strings: { content },
+    config: markdownlintConfig,
+    resultVersion: 3,
```
DavidAnson, line 58: Unnecessary, current default.

helpers/README.md
```diff
@@ -42,6 +42,29 @@ module.exports = {
...
+  });
+  const fixes = fixResults.content.filter((error) => error.fixInfo);
```
DavidAnson, line 60: Filtering should not be necessary, because I think violations without fixes are ignored. For simplicity, always directly calling the method should be fine.

DavidAnson (changes requested):
Thank you for the contribution! I left a few comments intended to simplify the example as much as possible. And a couple of purely stylistic things. Per my comment above, please replace to "next" branch. Thank you!

README.md
```diff
@@ -868,6 +868,10 @@ bad.md: 1: MD041/first-line-heading/first-line-h1 First line in a file should be
  Use --force to continue.
 ```
 
+#### Fixing violations
```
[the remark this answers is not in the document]
bmish: Oof, how embarrassing having a markdownlint violation in my markdownlint PR! Fixed.

README.md
```diff
@@ -868,6 +868,10 @@ bad.md: 1: MD041/first-line-heading/first-line-h1 First line in a file should be
...
+
+See [markdownlint-rule-helpers](./helpers/README.md#fixing-violations) for an example of how to use the Node API to auto-fix violations in a string.
```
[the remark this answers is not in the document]
bmish: Updated.

helpers/README.md
```diff
@@ -42,6 +42,29 @@ module.exports = {
 };
 ```
 
+### Fixing violations
```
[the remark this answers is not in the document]
bmish: Updated.

helpers/README.md
```diff
@@ -42,6 +42,29 @@ module.exports = {
...
+
+An example of how to use the Node API to auto-fix violations in a string:
+
+```js
+const { sync } = require('markdownlint');
+const markdownlintRuleHelpers = require('markdownlint-rule-helpers');
+const markdownlintConfig = require('./.markdownlint.json');
```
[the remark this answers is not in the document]
bmish: Removed.

helpers/README.md
bmish: Updated.

helpers/README.md
bmish: Removed.

helpers/README.md
bmish: Simplified.

helpers/README.md
```diff
@@ -42,6 +42,29 @@ module.exports = {
 };
 ```
 
+### Fixing violations
+
+An example of how to use the Node API to auto-fix violations in a string:
```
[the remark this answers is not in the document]
bmish: Removed.

helpers/README.md
bmish: I'm glad it's unnecessary. I was going to suggest that markdownlint needed a simpler API for fixing violations in a string (or file) without the need to have a bunch of logic/filtering for it. 

Calling `markdownlintSync` and then just `applyFixes` is much better.

DavidAnson (approved):
Looks great, thanks for the edits! There's a line length issue, but I may be able to fix that from my phone...

README.md
```diff
@@ -928,6 +928,10 @@ bad.md: 1: MD041/first-line-heading/first-line-h1 First line in a file should be
 
 <!-- markdownlint-restore -->
 
+### Fixing
+
+Rules that can be fixed automatically include a `fixInfo` property which is outlined in the [documentation for custom rules](./doc/CustomRules.md). To apply those fixes more easily, the `applyFixes` method in [markdownlint-rule-helpers](./helpers/README.md#applying-recommended-fixes) may be used.
```
bmish, line 933: Fix line length:
```suggestion
Rules that can be fixed automatically include a `fixInfo` property which is outlined in the [documentation for custom rules](./doc/CustomRules.md).
To apply those fixes more easily, the `applyFixes` method in [markdownlint-rule-helpers](./helpers/README.md#applying-recommended-fixes) may be used.
```

DavidAnson:
Thanks. I'll fix the new issue in a bit. There's a check of headings that needs an update.

DavidAnson:
Fixed in `next` via https://github.com/DavidAnson/markdownlint/commit/2c38ae90fc9794093370e5908c7555c086c08b38, thanks again!

Issue #721, opened 2023-02-15
YAML schema build can not handle arrays with multiple items


aepfli:
## issue

If there is a configuration parameter, which is an array of elements, and you have multiple default elements, it is not generating valid yaml

### input
```
     "defaults": [ "a", "b", "c"]
```

### actual

```yaml
defaults: [
   "a"
   "b"
   "c"
]
```

### expected

```yaml
defaults: 
   - "a"
   - "b"
   - "c"
```

## Proposal

By using `YAML.stringify` we have the appropriate behavior, with the usage of https://www.npmjs.com/package/yaml (additional dependency). Instead of doing replacements within the JSON string to currently generate the YAML.

PR #722, opened 2023-02-15
Improved YAML Schema build (#721)

aepfli:
Fixed issue with generating YAML schema by using
`YAML.stringify` instead of string replacements
of the `JSON.stringigy` result.

closes: #721

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,7 +88,8 @@
     "tv4": "1.3.0",
     "typescript": "4.9.4",
     "webpack": "5.75.0",
-    "webpack-cli": "5.0.1"
+    "webpack-cli": "5.0.1",
+    "yaml": "^2.2.1"
   },
   "keywords": [
     "markdown",
--- a/package.json
+++ b/package.json
@@ -89,7 +89,7 @@
     "typescript": "4.9.4",
     "webpack": "5.75.0",
     "webpack-cli": "5.0.1",
-    "yaml": "^2.2.1"
+    "yaml": "2.2.1"
   },
   "keywords": [
     "markdown",
--- a/schema/.markdownlint.yaml
+++ b/schema/.markdownlint.yaml
@@ -257,6 +257,5 @@ MD052: true
 # MD053/link-image-reference-definitions - Link and image reference definitions should be needed
 MD053:
   # Ignored definitions
-  ignored_definitions: [
-    "//"
-  ]
+  ignored_definitions:
+    - "//"
--- a/schema/.markdownlint.yaml
+++ b/schema/.markdownlint.yaml
@@ -1,4 +1,4 @@
-# Example markdownlint YAML configuration with all properties set to their default value
+# Example markdownlint configuration with all properties set to their default value
 
 # Default state for all rules
 default: true
--- a/schema/build-config-example.js
+++ b/schema/build-config-example.js
@@ -5,6 +5,7 @@
 const fs = require("node:fs");
 const path = require("node:path");
 const configSchema = require("./markdownlint-config-schema.json");
+const YAML = require("yaml");
 
 const configExample = {};
 for (const rule in configSchema.properties) {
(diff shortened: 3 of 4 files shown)


package.json
```diff
@@ -88,7 +88,8 @@
     "tv4": "1.3.0",
     "typescript": "4.9.4",
     "webpack": "5.75.0",
-    "webpack-cli": "5.0.1"
+    "webpack-cli": "5.0.1",
+    "yaml": "^2.2.1"
```
DavidAnson, line 92: Please use specific version (no ^) as with other packages.

schema/.markdownlint.yaml
```diff
@@ -257,6 +257,5 @@ MD052: true
 # MD053/link-image-reference-definitions - Link and image reference definitions should be needed
 MD053:
   # Ignored definitions
-  ignored_definitions: [
-    "//"
-  ]
\ No newline at end of file
+  ignored_definitions:
+    - "//"
```
DavidAnson, line 261: Yay!

schema/build-config-example.js
```diff
@@ -5,6 +5,7 @@
 const fs = require("node:fs");
 const path = require("node:path");
 const configSchema = require("./markdownlint-config-schema.json");
+const YAML = require("yaml");
```
DavidAnson, line 8: "yaml", please.

DavidAnson: Move one line up, please.

schema/build-config-example.js
```diff
@@ -36,18 +37,17 @@ fs.writeFileSync(
...
-  .replace(/,\n/g, "\n")
-  .replace(/^\{\n/, "")
-  .replace(/\}$/, "");
+const configStringYaml = YAML.stringify(configExample, {
+  "lineWidth": 0,
+  "defaultStringType": "QUOTE_DOUBLE",
+  "defaultKeyType": "PLAIN"
+})
+  .replace(/^(\s*)[^-\s]+-sub-description: "(.+)"$/gm, "$1# $2")
+  .replace(/^[^-\s]+-description: "(.+)"$/gm, "\n# $1");
```
DavidAnson, line 46: Can these two be combined? Seems the only difference is spaces and "-sub"? How does this look before transformation?

DavidAnson (changes requested):
Much cleaner, thank you! I left a handful of minor comments.

DavidAnson:
The linkcheck CI failure will go away when you update the PR to be for "next" branch instead of "main". (See CONTRIBUTING.md)

schema/build-config-example.js
aepfli: This is close to being the same regex used a few lines above for the JSON.

the description should add a new line before
sub-description should keep indentation but no new line.

I can try to unify but only focused on being consistent with the JSON file approach

DavidAnson: Okay, thanks for pointing that out! I'd forgotten what I did here. :| Looking again now, two thoughts:
1. What about using the same RegExp for JSON and YAML? They're operating on the same data, so that should be straightforward? It'd turn into something like `replace(descriptionRe, "json or yaml replacement text"` (and again for sub-description).
2. Can YAML use the same trick for creating the first line as JSON does? I was going to comment on it being a little unusual to do the string concat in the call to writeFileSync and this change would make the two more consistent and remove the need for that.

schema/build-config-example.js
```diff
@@ -36,18 +37,17 @@ fs.writeFileSync(
...
-  .replace(/\n {2}/g, "\n")
-  .replace(/\/\/ /g, "# ")
-  .replace(/(\s*)"([^"]+)":/g, "$1$2:")
-  .replace(/: \{/g, ":")
-  .replace(/\n\}/g, "")
-  .replace(/,\n/g, "\n")
-  .replace(/^\{\n/, "")
-  .replace(/\}$/, "");
+const configStringYaml = yaml.stringify(configExample, {
+  "lineWidth": 0,
```
DavidAnson, line 41: Are all 3 of these customizations necessary to produce the same output as today? If not, I prefer accepting the default vs. being explicit.

DavidAnson (changes requested):
Thanks for the clarification - morphed my feedback based on that.

schema/build-config-example.js
aepfli: - `lineWidth` is used for not breaking lines, this makes the comment transformation really hard

- the other two, are mainly to not configure double_quotes for string but leave it as plain for the keys
```
  # Heading style (without the setting)
  style: consistent
```
vs
```
  # Heading style (with the setting)
  style: "consistent"
```

regarding the quotes, this is easy to handle, the other one I would definitely keep.

schema/build-config-example.js
aepfli:
```
const commentify = (input, commentLiterals) => commentLiterals +
  // eslint-disable-next-line max-len
  " Example markdownlint configuration with all properties set to their default value\n" +
  input
    // eslint-disable-next-line max-len
    .replace(/^(\s*)"?[^-\s]+-sub-description"?: "?([^"\n]+)"?,?$/gm, "$1" + commentLiterals + " $2")
    // eslint-disable-next-line max-len
    .replace(/^(\s*)"?[^-\s]+-description"?: "?([^"\n]+)"?,?$/gm, "\n$1" + commentLiterals + " $2");
```

something like this?

aepfli:
thank you! - it was a pleasure

PR #868, opened 2023-06-10
# `MD033` - Inline HTML  Tags: `html`  Aliases: `no-inline-html`  Parameters:  - `allowed_elements`: Allowed elements (`string[]`, default `[]`)  This rule is triggered whenever raw HTML is used in a Markdown document:  ```markdown <h1>Inline HTML heading</h1> ```  To fix this, use 'pure' Markdown instead of including raw HTML:  ```markdown # Markdown heading ```  Note: To allow specific HTML elements, use the `allowed_elements` parameter.  Rationale: Raw HTML is allowed in Markdown, but this rule is included for those who want their documents to only include "pure" Markdown, or for those who are rendering Markdown documents into something other than HTML.Delete md033.md

Warda8841:


Warda8841:
Done

Warda8841:
Warda8841:patch-1

Warda8841:
Done

Issue #992, opened 2023-10-04
MD045: Should `no-alt-text` support HTML syntax?


khiga8:
Hi! Our team is really benefitting from the `no-alt-text` rule, but we noticed that it only supports inline image syntax.

Would you be open to supporting HTML img syntax?

PR #993, opened 2023-10-05
Add HTML support for MD045: no-alt-text

khiga8:
Issue: https://github.com/DavidAnson/markdownlint/issues/992

This PR adds support to flag missing alt text on HTML images. I wasn't sure if I should hold off on opening the PR until there was more discussion on the issue so this might be premature... I am happy to continue discussing this on this PR or on the issue.

#### Approach taken

I am new to this repo and followed the approaches I saw in other rules.

I noticed there is parser that returns HTML tokens. I didn't see a way to access the attributes off the HTML. It may not be ideal but I ended up using regex to check if the HTML image token contains `alt=`.


DavidAnson:
This looks pretty thorough and I like the test cases I see. Thank you! I will need to review this from a real computer, potentially this evening. I will have a set of nitpicks as always, so be warned. :)

A few things I noticed from a very quick scan on my phone:
- Element and attribute comparisons should be case insensitive, this suggests there are no test cases for uppercase
- I do not understand how range can be calculated after removing line breaks, this may mean there is something of a gap in the tests
- If you convert this to a real PR, we can see the result of CI, you can get close locally with "npm run ci"
- There are three copies of the documentation and you modified the correct one, but did not run the npm script to update the other two
- I include a space between rule names in test files {MD001} {MD002}
- I might choose to add the alternate text tag to the couple of unrelated test files instead of suppressing this rule as you did

khiga8:
Hi @DavidAnson, I appreciate your feedback! I've addressed the following from your first set of feedback:

> Element and attribute comparisons should be case insensitive, this suggests there are no test cases for uppercase

> There are three copies of the documentation and you modified the correct one, but did not run the npm script to update the other two

> I include a space between rule names in test files {MD001} {MD002}

>I might choose to add the alternate text tag to the couple of unrelated test files instead of suppressing this rule as you did

With regards to: 
>I do not understand how range can be calculated after removing line breaks, this may mean there is something of a gap in the tests

I followed the approach taken in [MD033.js](https://github.com/DavidAnson/markdownlint/blob/main/lib/md033.js) which was one of the only other rules I could find that sets an error on an HTML tag. I believe this sets the error range on the first line of the tag and accounts for multiline tags. My understanding could be wrong and I am definitely open to your suggestions.

DavidAnson:
Thanks for the quick activity! I will remind myself how the other range calculation works when I review this.

DavidAnson:
FYI, GitHub is complaining this PR cannot be merged due to conflicts and I think that may be blocking me from approving a CI run?

DavidAnson:
FYI now that tests are running, if anything new pops up in TestRepos, I typically verify all new issues are valid before updating those snapshots. There are npm scripts to clone and lint and update.

test/snapshots/markdownlint-test-repos.js.md
```diff
@@ -8,7 +8,7 @@ Generated by [AVA](https://avajs.dev).
 
 > Expected linting violations
 
-    ''
+    'test-repos/apache-airflow/tests/system/providers/apache/hive/example_twitter_README.md: 36: MD045/no-alt-text Images should have alternate text (alt text) [Element: img]'
```
DavidAnson, line 11: Fewer than I thought! Please let me know if you did not visually verify each one of these and I will.

DavidAnson (changes requested):
Looks solid, thank you! Most of my comments are trivial nitpicks.

lib/md045.js
```diff
@@ -26,5 +28,26 @@ module.exports = {
         );
       }
     }
+
+    const html = filterByTypes(params.parsers.micromark.tokens, [ "htmlText" ]);
+    for (const token of html) {
+      const htmlTagInfo = getHtmlTagInfo(token);
+      if (htmlTagInfo && htmlTagInfo.name.toLowerCase() === "img") {
+        const hasAltAttribute = new RegExp(/alt=/i);
+        if (!hasAltAttribute.test(token.text)) {
```
iansan5653, line 37: Looks like this will only error if the alt text is missing entirely. Should we also error if the alt text is empty (`alt=""`)? Or is that technically OK since it could indicate a presentational image?

DavidAnson: I suggest it is reasonable as implemented because it is clear in that case that the author considered what the alternate link should be. However, the existing implementation looks like it will flag empty string which would be inconsistent. However, if I recall correctly (on phone currently) that may be due to how the parser represents this rather than a deliberate decision to flag present-but-empty alternate text.

test/snapshots/markdownlint-test-repos.js.md
```diff
@@ -37,7 +37,13 @@ Generated by [AVA](https://avajs.dev).
...
 > Expected linting violations
 
-    `test-repos/mdn-content/files/en-us/web/svg/attribute/d/index.md: 234: MD007/ul-indent Unordered list indentation [Expected: 0; Actual: 10]␊
+    `test-repos/mdn-content/files/en-us/web/css/angle/index.md: 38: MD045/no-alt-text Images should have alternate text (alt text)␊
+    test-repos/mdn-content/files/en-us/web/css/angle/index.md: 49: MD045/no-alt-text Images should have alternate text (alt text)␊
+    test-repos/mdn-content/files/en-us/web/css/angle/index.md: 60: MD045/no-alt-text Images should have alternate text (alt text)␊
+    test-repos/mdn-content/files/en-us/web/css/angle/index.md: 71: MD045/no-alt-text Images should have alternate text (alt text)␊
+    test-repos/mdn-content/files/en-us/web/css/transform-function/rotate3d/index.md: 67: MD045/no-alt-text Images should have alternate text (alt text)␊
+    test-repos/mdn-content/files/en-us/web/css/transform-function/rotate3d/index.md: 251: MD045/no-alt-text Images should have alternate text (alt text)␊
+    test-repos/mdn-content/files/en-us/web/svg/attribute/d/index.md: 234: MD007/ul-indent Unordered list indentation [Expected: 0; Actual: 10]␊
```
khiga8, line 46: This is an unrelated rule violation. Not sure if it's expected.

khiga8:
Hi @DavidAnson! Apologies for the delayed follow-up.

Starting from commit https://github.com/DavidAnson/markdownlint/pull/993/commits/0165cf569e19895ff0eab385dd68761ca7e0a294, I've hopefully addressed all of your feedback, but let me know if I'm missing anything.


Re: https://github.com/DavidAnson/markdownlint/pull/993#discussion_r1348216640
> Please let me know if you did not visually verify each one of these and I will.


I've visually verified the repo snapshot flags with the following steps:
1. In my VSCode editor, I did a regex search of `<img` and filtered to `test-repos/**/*.md`. This gives me 19 results across 5 files. 
2. In the snapshots, there are 9 new image violations being flagged.
3. I've verified that each of the 9 flagged violations are legit violations. I saw that the remaining 10 images all have alt text on it.

lib/md045.js
khiga8: Yes, my intention is for this rule to allow decorative images (set via empty strings).

test/snapshots/markdownlint-test-repos.js.md
DavidAnson: If you're fully synced, this should already be accounted for in the test.

DavidAnson:
Thank you for verifying the new violations - that's exactly what I wanted.

helpers/helpers.js
```diff
@@ -5,7 +5,10 @@
 const micromark = require("./micromark.cjs");
 
 const { newLineRe } = require("./shared.js");
+const { nextLinesRe } = require("./shared.js");
```
DavidAnson, line 8: This and the previous line can be merged into one. I may push this to your branch myself.

lib/md045.js
```diff
@@ -2,15 +2,19 @@
 
 "use strict";
 
-const { addError } = require("../helpers");
-const { filterByTypes } = require("../helpers/micromark.cjs");
+const { addError, nextLinesRe } = require("../helpers");
+const { filterByTypes, getHtmlTagInfo } = require("../helpers/micromark.cjs");
+
+// Regular expression for identifying alt attribute
+const altRe = /\salt=/i;
```
DavidAnson, line 9: For comparison, I use this elsewhere for similar purpose: `const idRe = /\sid\s*=\s*['"]?([^'"\s>]+)/iu;`. I'd like to try that here.

DavidAnson: Actually, there should maybe be a getHtmlAttributeRe(name) function for this.

lib/md045.js
```diff
@@ -2,15 +2,19 @@
...
 
 module.exports = {
   "names": [ "MD045", "no-alt-text" ],
   "description": "Images should have alternate text (alt text)",
   "tags": [ "accessibility", "images" ],
   "function": function MD045(params, onError) {
-    const images = filterByTypes(params.parsers.micromark.tokens, [ "image" ]);
+    const tokens = params.parsers.micromark.tokens;
```
DavidAnson, line 16: Not wrong, but for consistency: `const { tokens } = ...`

lib/md045.js
```diff
@@ -26,5 +30,24 @@ module.exports = {
         );
       }
     }
+
+    const htmls = filterByTypes(tokens, [ "htmlText" ]);
```
DavidAnson, line 34: Note to self: `htmlTexts` and `htmlText`.

lib/md045.js
```diff
@@ -26,5 +30,24 @@ module.exports = {
...
+    for (const html of htmls) {
+      const htmlTagInfo = getHtmlTagInfo(html);
+      const isImage = htmlTagInfo && htmlTagInfo.name.toLowerCase() === "img";
```
DavidAnson, line 37: Parenthesize the clause after the `&&`, please.

lib/md045.js
```diff
@@ -26,5 +30,24 @@ module.exports = {
...
+      if (isImage && !altRe.test(html.text)) {
```
DavidAnson, line 38: Let's merge the 3 conditions into one comparison.

lib/md045.js
```diff
@@ -26,5 +30,24 @@ module.exports = {
...
+        const range = [
+          html.startColumn,
```
DavidAnson, line 40: I might capture all 3 `html.` properties for reuse (for consistency).

DavidAnson (changes requested):
Looks awesome. I'm going to make the changes I suggest since I'm getting extra nit-picky at this point.

lib/md045.js
DavidAnson: I'll implement this separately as it's a little more involved.

DavidAnson (approved):
GitHub is being silly, but maybe you can tell me how to work around it. If I make edits and push changes from a Codespace, it fails with HTTP 403 because I do not have write permission to your fork. However, if I use the web UI to make edits to files you have already edited, it works (that's what I did to implement my feedback), BUT it always adds a trailing newline, so the `git diff` check fails on the generated file. Could you please sync my changes to your branch, re-generate `markdownlint-browser.js` in your fork, and push it? Everything should go green and I'll approve the PR. (OR maybe you know how I can push edits from a Codespace?)

khiga8:
Thanks, I've pushed up a commit(https://github.com/DavidAnson/markdownlint/pull/993/commits/a1db76d26fcbe4462cf9bc31b0010508349bb586) which re-runs the script!

> GitHub is being silly, but maybe you can tell me how to work around it. If I make edits and push changes from a Codespace, it fails with HTTP 403 because I do not have write permission to your fork. 

That sounds frustrating! I do have this checkbox option toggled on this PR:
Screenshot of the 'Allow edits and access to secrets by maintainers' checkbox on this PR toggled

This checkbox  should give you permission to commit to my branch. (Related: [Allowing changes to a pull request branch created from a fork](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork))

Is your codespaces of my forked repo? This article has some pointers that might be helpful: [Committing changes to a pull request branch created from a fork - GitHub Docs](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/committing-changes-to-a-pull-request-branch-created-from-a-fork).

DavidAnson:
Thank you! I will do a final review and merge the PR this evening.

Regarding permissions, I think all the preconditions are correct (and my ability to change individual files confirms my permission), but pushing from two different Codespaces created from that pull request failed. I have tried this in the past from the command line and that failed as well. It is possible I am making a mistake every time, but this has never worked for me and I am sure that I created the Codespace from the PR yesterday, so would have expected GitHub to set things up correctly. (I confirm that it created an "origin" and "upstream" remote.) I will try again tonight following the directions you sent exactly to double check myself.

khiga8:
> It is possible I am making a mistake every time, but this has never worked for me and I am sure that I created the Codespace from the PR yesterday,

@DavidAnson I was able to repro what you're seeing on my end too. After some investigation, I discovered this is a known issue in Codespaces currently.

Related community issues:
* https://github.com/orgs/community/discussions/36357
* https://github.com/orgs/community/discussions/36299

For now, you will have to stick with the traditional route of cloning via the steps in [Committing changes to a pull request branch created from a fork - GitHub Docs](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/committing-changes-to-a-pull-request-branch-created-from-a-fork). 😅

DavidAnson:
Merged, thank you!

Regarding the Codespaces issue, thanks for tracking that down. :)

PR #1011, opened 2023-10-17
Enhancement: MD007 parameters order

Malix-off:
_nit, but still_


DavidAnson:
Parameters are listed in alphabetical order (by a script) so this is correct as-is.

Issue #994, opened 2023-10-06
Make helpers in `micromark.cjs` available through `markdownlint-rule-helpers`


khiga8:
Our team is writing some custom rules for our projects. We would like to test these rules in our projects and upstream it here, if it makes sense.

I notice there are helpful methods available as an NPM package to support writing custom lint rules in [markdownlint-rule-helpers](https://github.com/DavidAnson/markdownlint/tree/main/helpers). 

I'm hoping to be able to use the `getHtmlTagInfo` and `filterByTypes` functions which are currently in `micromark.cjs`, but these don't appear to be available as part of the package.

I'd like to avoid repro-ing/copying this logic into our project -- would you be open to making these micromark functions available as part of [markdownlint-rule-helpers](https://github.com/DavidAnson/markdownlint/tree/main/helpers)?

DavidAnson:
Quick history lesson for context:
- The `markdownlint-rule-helpers` package is unsupported and provided for convenience only
- Anything in that package related to the parser is going to work with the `markdown-it` parser
- I've been gradually migrating everything over to the `micromark` parser (which provides rich positional information) and didn't want to introduce any public dependencies on that until I felt more confident about what I'm doing 
- Once migrated, I hope to stop running the `markdown-it` parser except in the presence of custom rules which so far only have access to it (officially)
- I may even stop publishing `markdownlint-rule-helpers`
- Undocumented access to both parsers has been available to custom rules for a while now
- As recently as the current version, I am still tweaking some of the helpers in `micromark.cjs`

So I'm not sure I'd like to go down the road you outline... I'm open to suggestions, but propose that copying those two functions is reasonable at the current time.

PR #1015, opened 2023-10-21
Add missing `parsers` property to `RuleParams`

Lehoczky:
## 📚Description

I've found that `RuleParams` type was missing the `parsers` property while playing around with the code.

What do you think about using type hints in rule declarations [like it is done in ESLint](https://github.com/eslint/eslint/blob/main/lib/rules/array-bracket-newline.js#L14)? It would elevate the developer experience when working with rules not just because of the type checking, but because of the added autocompletion and better syntax highlighting:

image
image


DavidAnson:
This is deliberately omitted and undocumented for now. More context in #994.

Lehoczky:
Oh okay.

PR #1019, opened 2023-10-23
Fix super-linter URL

koppor:
The linked super-linter repo is a fork only:

image

This PR updates the URL to point to the main repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ and test cases came directly from that project.
 [nodejs-extensions]: https://github.com/Lombiq/NodeJs-Extensions
 [rubygems-mdl]: https://rubygems.org/gems/mdl
 [sublimelinter]: https://packagecontrol.io/packages/SublimeLinter-contrib-markdownlint
-[super-linter]: https://github.com/github/super-linter
+[super-linter]: https://github.com/super-linter/super-linter
 [vscode-markdownlint]: https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint
 
 ### References


DavidAnson:
Please update PR to apply to "next" branch.

koppor:
> Please update PR to apply to "next" branch.

Thank you for retargeting. I also merged `next` to be sure that no conflicts occur.

DavidAnson:
I do not want a merge commit and I do not know how GitHub UI will handle this, so I may need to do some command line tweaking to manually accept this change.

koppor:
> I do not want a merge commit and I do not know how GitHub UI will handle this, so I may need to do some command line tweaking to manually accept this change.

Sure thing. GitHub UI should offer you "Squash and merge" - then you get a single commit containing the change here with a reference to the Pull Request (using `#1019`).

Example (from other repositories)

image

At https://github.com/DavidAnson/markdownlint/settings you can even forbid merge commits in general :p

DavidAnson:
My fear is that squash and merge will lead to a single commit that is also a merge. I've used it before, but not in a merge scenario like this. The problem is I won't know until it's too late because main and next are both protected branches in GitHub.

koppor:
First of all, thank you for merging!

> My fear is that squash and merge will lead to a single commit that is also a merge. I've used it before, but not in a merge scenario like this. The problem is I won't know until it's too late because main and next are both protected branches in GitHub.

I can fully follow. Each time, I press "squash and merge" at my JabRef repo, I have the same feeling. However, it worked for years now... -- The feeling when using merge queues gets worse - but we still have them in place :p

PR #1043, opened 2023-11-17
new table border rule (fixes #93)

### LINKED ISSUE #90: Table alignment checks
This could produce an alignment warning, enticing people to draw text-clean tables:

```
aze | qsd
------|----
123 | 456
```

### LINKED ISSUE #93: MD055: Table style consistency check
It would be nice to be able to enforce table style consistency:

```
| foo | bar |
|-----|-----|
| 123 | 456 |
```

vs

```
| foo | bar
|-----|----
| 123 | 456
```

vs

```
foo | bar
----|----
123 | 456
```

Style could be detected as being inconsistent from first occurence, or globally enforced as a setting.

gabriel-check24:


README.md
```diff
@@ -175,6 +176,7 @@ rules at once.
 - **`ol`** - `MD029`, `MD030`, `MD032`
 - **`spaces`** - `MD018`, `MD019`, `MD020`, `MD021`, `MD023`
 - **`spelling`** - `MD044`
+- **`tables`** - `MD055`
```
DavidAnson, line 179: "table", please. (I know this list is not super consistent.)

README.md
```diff
@@ -134,6 +134,7 @@ playground for learning and exploring.
 - **[MD052](doc/md052.md)** *reference-links-images* - Reference links and images should use a label that is defined
 - **[MD053](doc/md053.md)** *link-image-reference-definitions* - Link and image reference definitions should be needed
 - **[MD054](doc/md054.md)** *link-image-style* - Link and image style
+- **[MD055](doc/md055.md)** *table-missing-border* - Table is missing border pipes
```
DavidAnson, line 137: I may want to say "Table is missing leading or trailing pipe character" based on <https://github.github.com/gfm/#tables-extension->.

doc-build/md055.md
```diff
@@ -0,0 +1,20 @@
+This rule is triggered when a table is missing a border. Borders consist of the
```
DavidAnson, line 1: Similar comment as above, maybe something like "... triggered when a table is missing a leading or trailing pipe character."

doc-build/md055.md
```diff
@@ -0,0 +1,20 @@
...
+pipe '|' character.
```
DavidAnson, line 2: Use backticks instead of single quotes, please, as this is a literal.

doc-build/md055.md
```diff
@@ -0,0 +1,20 @@
...
+
+This table is missing a border on the both sides:
+
+```markdown
+| Heading | Heading |
+|---------|---------
+  Data    | Data    |
+```
+
+To fix this, add a pipe character at the start or end of the row.
```
DavidAnson, line 12: End with ":"

doc-build/md055.md
```diff
@@ -0,0 +1,20 @@
+This rule is triggered when a table is missing a border. Borders consist of the
+pipe '|' character.
+
+This table is missing a border on the both sides:
```
DavidAnson, line 4: "is missing a trailing pipe on its second line and a leading pipe on its third line"

doc-build/md055.md
```diff
@@ -0,0 +1,20 @@
...
+
+```markdown
+| Heading | Heading |
+|---------|---------
+  Data    | Data    |
```
DavidAnson, line 9: Nitpick: "Data" -> "Cell"

doc-build/md055.md
```diff
@@ -0,0 +1,20 @@
...
+
+To fix this, add a pipe character at the start or end of the row.
+
+```markdown
+| Heading | Heading |
+|---------|---------|
+| Data    | Data    |
+```
+
+Rationale: Some renderers may render tables without borders incorrectly.
```
DavidAnson, line 20: "some parsers may have difficult with tables that are missing their leading or trailing pipe characters"

DavidAnson, line 20: Is this true, by the way? Do you have any examples of scenarios where this breaks parsing? (I think it IS true, but couldn't say what breaks.)

lib/rules.js
```diff
@@ -51,8 +51,8 @@ const rules = [
   require("./md051"),
   require("./md052"),
   require("./md053"),
-  require("./md054")
-  // md055: See https://github.com/markdownlint/markdownlint
```
DavidAnson, line 55: Nice!

schema/build-config-schema.js
```diff
@@ -531,6 +531,20 @@ for (const rule of rules) {
         },
       };
       break;
+    case "MD055":
+      scheme.properties = {
+        "style": {
+          "description": "Where border pipes are required",
```
DavidAnson, line 537: Why is this optional? My first thought is to have this rule require both - or else turn off the rule to require neither or not care. Supporting one-or-the-other is more code and more testing. Would anyone customize this?

test/break-all-the-rules.md
```diff
@@ -105,6 +105,11 @@ Strong **with** different style {MD050}
 <!-- markdownlint-disable-next-line MD053 -->
 [url]: https://example.com/page
 
+| table | text |
```
DavidAnson, line 108: Wow, good catch, most folks miss this file. :) To be a nitpicker, just a single violation is all that's needed.

test/html-comment-in-markdown-table.md
```diff
@@ -51,3 +51,7 @@
 | cell  | <!--           \
 | cell  | `{MD038} ` --> |
 | cell  | cell           |
+
+<!-- markdownlint-configure-file {
+  "MD055": false
```
DavidAnson, line 56: Good. As of a couple of weeks ago, the standard is to use the rule name instead of its number when configuring like this.

test/long_lines_code.md
```diff
@@ -29,6 +29,7 @@ This is a short line.
 | Content Cell   | Content Cell   | Content Cell   | Content Cell   | Content Cell   | Content Cell   |
 | ============== | ============== | ============== | ============== | ============== | ============== |
 | Footer Cell    | Footer Cell    | Footer Cell    | Footer Cell    | Footer Cell    | Footer Cell    |
+
 {: rules="groups"}
```
DavidAnson, line 33: If you don't know what this syntax is trying to do, let's maybe delete it?

test/md033-cannot-read-property-of-undefined.md
```diff
@@ -9,3 +9,7 @@ Head    | Head
 <br/>   | {MD033}
 text    | text
 {MD033} | <b>text</b>
+
+<!-- markdownlint-configure-file {
+  "table-missing-border": false
```
DavidAnson, line 14: Yes, "This is the way." :)

test/md055-missing-pipes.md
```diff
@@ -0,0 +1,19 @@
+# MD055 Missing Borders
```
DavidAnson, line 1: This is great. I'd like to see some multi-row tables, too, with missing pipes on the first/middle/last rows. Let's be sure the line numbers are correct.

DavidAnson, line 1: Please add tests for missing pipe on the header line and the separator line, too.

DavidAnson, line 1: While there are exceptions, I prefer test files not include rule names. Something like `table-missing-pipes.md` would be good here. Also, the Heading text should match the file name (here "pipes" vs. "borders").

lib/md055.js
```diff
@@ -0,0 +1,40 @@
+// @ts-check
+
+"use strict";
+
+const { addError} = require("../helpers");
```
DavidAnson, line 5: Missing ' ' before '}'

lib/md055.js
```diff
@@ -0,0 +1,40 @@
...
+
+const noLeadingPipeErrorMessage = "Table rows must start with a pipe";
+const noTrailingPipeErrorMessage = "Table rows must end with a pipe";
+
+module.exports = {
+    "names": ["MD055", "table-missing-border"],
+    "description": "Table is missing border pipes",
+    "information": new URL("https://github.github.com/gfm/#tables-extension-"),
```
DavidAnson, line 13: Please don't override this field. But it's a good callout that the GFM link should be added to the rule documentation, please!

lib/md055.js
```diff
@@ -0,0 +1,40 @@
...
+    "tags": ["tables"],
```
DavidAnson, line 14: Missing ' ' inside '[' and ']'.

lib/md055.js
```diff
@@ -0,0 +1,40 @@
...
+    "function": function MD055(params, onError) {
+        const outerPipes = String(params.config.style || "both");
```
DavidAnson, line 16: This will simplify if we remove the parameter.

lib/md055.js
```diff
@@ -0,0 +1,40 @@
...
+        const needLeading = outerPipes === "both" || outerPipes === "leading";
+        const needTrailing = outerPipes === "both" || outerPipes === "trailing";
+
+        const tables = params.parsers.markdownit.tokens.filter(
```
DavidAnson, line 20: Please use micromark tokens instead - I'm trying to remove all use of `markdown-it`.

lib/md055.js
```diff
@@ -0,0 +1,40 @@
...
+            (token) => token.type === "table_open"
+        );
+
+        for (const table of tables) {
```
DavidAnson, line 24: Skipping this since it'll change for micromark tokens.

lib/md055.js
```diff
@@ -0,0 +1,40 @@
...
+            const [tableStart, tableEnd] = table.map;
+            const lines = params.lines;
+            const indentation = params.lines[tableStart].indexOf("|");
```
DavidAnson, line 27: But good news with micromark is that you shouldn't have to do unreliable line parsing like this as there tend to be tokens for everything.

lib/md055.js
```diff
@@ -0,0 +1,40 @@
...
+            for (let line = tableStart; line < tableEnd; line++) {
+                const startsWithPipe = lines[line].charAt(indentation) === "|";
+                if (!startsWithPipe && needLeading) {
+                    addError(onError, line + 1, noLeadingPipeErrorMessage, lines[line]);
+                }
+
+                if (!lines[line].endsWith("|") && needTrailing) {
+                    addError(onError, line + 1, noTrailingPipeErrorMessage, lines[line]);
```
DavidAnson, line 35: There are helpers for `addError` that will automatically trim the context string when it's very long - please use them.

DavidAnson (changes requested):
Thanks for doing this! You figured out a lot of the things that trip people up. :) Two big issues I saw are using micromark tokens and deciding if we really need the leading/trailing parameter. Plus some more testing around missing pipes in the header/delimeter row. And please add some use of ":" for row alignment, though I don't think it'll break anything. And I had a bunch of nitpicks, sorry.

DavidAnson:
For the TestRepos failures, please verify every new issue is valid, then you can `update-snapshots-test-repos` them. If you didn't see, `clone-test-repos` will pull that stuff down locally and `lint-test-repos` will run those tests. They take about a minute to run.

DavidAnson:
The Checkers/spelling issue will go away with an edit I proposed above.

doc-build/md055.md
Friendly-Banana: Not many, but apparently StackOverflow has some issues [one line table](https://meta.stackexchange.com/questions/372897/incomplete-table-markdown-causes-preview-to-appear-fine-while-its-broken-when-p), [Missing column](https://meta.stackexchange.com/questions/371475/some-markdown-tables-incorrectly-omit-an-entire-column-in-the-rendered-post-but)

schema/build-config-schema.js
Friendly-Banana: #93 includes an example with a border on one side. I'd also go with removing this complexity though.

test/long_lines_code.md
Friendly-Banana: Sure, :smile:,

test/md055-missing-pipes.md
Friendly-Banana: ok, I'll try, but as mentioned already some of that won't be recognized as a table anymore.

doc-build/md055.md
gabriel-check24: Apparently StackOverflow has some issues [one line table](https://meta.stackexchange.com/questions/372897/incomplete-table-markdown-causes-preview-to-appear-fine-while-its-broken-when-p), [Missing column](https://meta.stackexchange.com/questions/371475/some-markdown-tables-incorrectly-omit-an-entire-column-in-the-rendered-post-but)
Also, IntelliJ doesn't render an inline preview when a pipe is missing.

test/long_lines_code.md
gabriel-check24: Sure, :smile:,

gabriel-check24:
Lots of feedback, thanks, working through it. 

The test repos are looking good, except for `v8-v8-dev/src/features/promise-combinators.md:29` . Here the parser is at fault, it categorizes the `:::` as belonging to the table. This issue, no blank line after a table, would be recognized by one of the other table rules btw :).  What should I do?

```
:::table-wrapper
| name                                        | description                                     | status                                                          |
| ------------------------------------------- | ----------------------------------------------- | --------------------------------------------------------------- |
| [`Promise.allSettled`](#promise.allsettled) | does not short-circuit                          | [added in ES2020 ✅](https://github.com/tc39/proposal-promise-allSettled) |
:::
```

Can you squash merge so we have a commit history here or should I already squash?

DavidAnson:
Based on my demo site which uses the micromark parser by default, it does not appear to me that the lines beginning with three colons are treated as part of the table: <https://dlaa.me/markdownlint/>. I need to look at this on a computer instead of my phone to understand what you are saying.

Regarding squashing commits, please feel free to do so. While it can be helpful sometimes during reviews to have new changes identifiable in separate commits, this will eventually go into next branch as a single commit.

DavidAnson:
(Please let me know when this PR is ready for the next review pass. No rush.)

gabriel-check24:
How do you use the Demo to determine what is parsed as a table? I can only see the finished markdown.

You can set a conditional breakpoint in `md055.js` line 22 with condition `row.text === ":::"` and then run `lint-test-repos`. This breakpoint gets triggered despite the earlier `.filter((token) => token.type === "tableRow" || token.type === "tableDelimiterRow")`.

gabriel-check24:
This issue is not specific to the v8 table-wrapper syntax, tables require a blank line to end for the parser. So e.g.
```
|table|demo|
|---|---|
|row|row|
hallo
```
doesn't work either.

Apart from that issue, I'm ready for the next review.

gabriel-check24:
Okay, so [this behaviour is wanted](https://github.com/orgs/micromark/discussions/160#discussioncomment-7660398), GitHub also behaves that way. I updated the snapshots to include this error.

micromark/exports-html.mjs
```diff
@@ -1,7 +1,5 @@
 // @ts-check
 
-/* eslint-disable n/no-missing-import */
```
DavidAnson, line 3: Please undo both instances of this edit. The disable is necessary in the general case and only unnecessary if someone runs `npm install` in the `micromark` directory. Removing like this will fail CI.

test/markdownlint-test.js
```diff
@@ -475,13 +475,14 @@ test("styleAll", (t) => new Promise((resolve) => {
...
         "MD046": [ 49, 73, 77 ],
-        "MD047": [ 126 ],
+        "MD047": [ 130 ],
         "MD048": [ 77 ],
         "MD049": [ 90 ],
         "MD050": [ 94 ],
         "MD051": [ 96 ],
         "MD052": [ 98 ],
-        "MD053": [ 100 ]
+        "MD053": [ 100 ],
```
DavidAnson, line 484: When you merge, there should be an entry for MD054 here.

test/long_lines_code.md
```diff
@@ -29,7 +29,6 @@ This is a short line.
 | Content Cell   | Content Cell   | Content Cell   | Content Cell   | Content Cell   | Content Cell   |
 | ============== | ============== | ============== | ============== | ============== | ============== |
 | Footer Cell    | Footer Cell    | Footer Cell    | Footer Cell    | Footer Cell    | Footer Cell    |
-{: rules="groups"}
```
DavidAnson, line 32: Thank you!

test/table-missing-pipes.md
```diff
@@ -0,0 +1,127 @@
+# Table Missing Pipes
```
DavidAnson, line 1: YES! Love how this file grew!

test/table-missing-pipes.md
```diff
@@ -0,0 +1,127 @@
...
+
+{MD055:17}
+
+| Table | Header |
+ -------|--------|
+
+{MD055:22}
+
+| Table | Header |
+ -------|---------
```
DavidAnson, line 27: I notice the missing start pipe is replaced with a space, but the missing end pipe is replaced with a dash - is that significant?

test/table-missing-pipes.md
```diff
@@ -0,0 +1,127 @@
...
+ {MD055} | cell
+| cell   | cell   |
+
+| Table  | Header |
+|--------|--------|
+| cell   | cell   |
+| cell   | cell   |
+ {MD055} | cell
+
+## No false-positive
```
DavidAnson, line 117: YAY!

test/table-missing-pipes.md
```diff
@@ -0,0 +1,127 @@
...
+
+| Table | Header |
+|-------|--------|
+| cell  | cell   |
+
+| Table | Header |
+|-------|--------|
+| cell  | cell   |
+| cell  | cell   |
+| cell  | cell   |
```
DavidAnson, line 127: Could I ask for one more `##` section in this file called "No trailing blank line" that includes an example of that scenario and triggers the rule per your investigation? It'll be something to confirm that behavior and maybe point people to if they ask.

doc-build/md055.md
```diff
@@ -0,0 +1,22 @@
...
+  Data    | Data    |
+```
+
+To fix this, add a pipe character at the start or end of the row:
+
+```markdown
+| Heading | Heading |
+|---------|---------|
+| Data    | Data    |
+```
```
DavidAnson, line 19: Please add after this something like:

```md
Note that text immediately below a table is treated as part of the table and will trigger this rule:

[EXAMPLE]
```

lib/md055.js
```diff
@@ -0,0 +1,28 @@
+// @ts-check
+
+"use strict";
+
+const { addErrorContext } = require("../helpers");
+const { filterByTypes } = require("../helpers/micromark.cjs");
+
+const noPipeErrorMessage = "Table row is missing a leading or trailing pipe: ";
```
DavidAnson, line 8: This message duplicates the name of the rule and can be removed. The context of a violation is usually just the violation, `row.text` as you have it below.

lib/md055.js
```diff
@@ -0,0 +1,28 @@
...
+
+module.exports = {
+    "names": [ "MD055", "table-missing-border" ],
+    "description": "Table is missing leading or trailing pipe character",
+    "tags": [ "table" ],
+    "function": function MD055(params, onError) {
+        const tables = filterByTypes(params.parsers.micromark.tokens, ["table"]);
+
```
DavidAnson, line 16: Nitpick, remove empty line, please. (Aside, I like that this code got smaller!)

lib/md055.js
```diff
@@ -0,0 +1,28 @@
...
+        for (const table of tables) {
+            const rows = table.children.flatMap((t) => t.children)
```
DavidAnson, line 18: Why not use `filterByTypes` again?

lib/md055.js
```diff
@@ -0,0 +1,28 @@
...
+                .filter((token) => token.type === "tableRow" || token.type === "tableDelimiterRow");
+
```
DavidAnson, line 20: No blank line.

lib/md055.js
```diff
@@ -0,0 +1,28 @@
...
+            for (const row of rows) {
+                if (!row.text.startsWith("|") || !row.text.endsWith("|")) {
```
DavidAnson, line 22: Love how clean and clear this is!

lib/md055.js
```diff
@@ -0,0 +1,28 @@
...
+                    addErrorContext(onError, row.startLine, noPipeErrorMessage + row.text);
```
DavidAnson, line 23: Please have a look at the use of `left` and `right` parameters to make sure truncation happens in the right place for long strings: <https://github.com/DavidAnson/markdownlint/blob/main/lib/md021.js>. And let's split this so it creates one error if the beginning is missing and another error if the end is missing. That'll make passing left/right easy.

AND that change sets up for the next step (which you should NOT do as part of this PR and I am happy to do if you do not want to) which is to add `fixInfo` so this rule can be automatically fixed. I think that will be straightforward much of the time, though maybe lining up the end pipe will be messy if someone tried to format their table.

DavidAnson (changes requested):
I skipped reviewing the snapshots yet, but this is looking great. Thanks so much for keeping at it! Per your investigation into trailing text, I’d like to call that out in the tests and the documentation. The rule implementation is nice and tight now which is awesome. :)

FYI, to resolve conflicts in snapshots, what I usually do is to start the rebase and then when git pauses to resolve the conflict, run the npm script to update, stage the result, and continue the rebase. I don’t think GitHub will offer to run CI for this PR until that’s resolved.

doc-build/md055.md
```diff
@@ -0,0 +1,22 @@
+This rule is triggered when
+a [table](https://github.github.com/gfm/#tables-extension-) is missing a leading
+or trailing pipe character `|`.
+
+This table is missing pipes on both sides:
```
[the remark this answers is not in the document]
gabriel-check24: the delimiter row is missing a pipe at the end:
> |---------|---------

I can change it its own row to be more obvious?

test/table-missing-pipes.md
gabriel-check24: Not significant afaik, it may be an artifact from my editor trying to fix the table. Should I change it for consistency?

test/table-missing-pipes.md
gabriel-check24: done

lib/md055.js
gabriel-check24: I split it. It'd be great if you can add the `fixInfo`.

DavidAnson:
Please let me know when I should take the next look.

gabriel-check24:
Please take a look and provide some clarification [here](https://github.com/DavidAnson/markdownlint/pull/1043#discussion_r1405851038) [here](https://github.com/DavidAnson/markdownlint/pull/1043#discussion_r1405844670) and [here](https://github.com/DavidAnson/markdownlint/pull/1043#discussion_r1404624425)

micromark/exports-html.mjs
DavidAnson: Yes. Or clean the micro ark directory and it should go away.

doc-build/md055.md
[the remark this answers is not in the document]
DavidAnson: Oh, sorry, my mistake. It only needs to be obvious to the computer. Leave it as-is.

test/table-missing-pipes.md
DavidAnson: If you don’t mind, please.

DavidAnson:
Clarification to those three things provided. To fix the “Checkers“ break, add “GFM“ to dictionary.txt in sorted order. I’ll do another review pass later tonight, probably.

doc-build/md055.md
```diff
@@ -0,0 +1,34 @@
...
+| Cell    | Cell    |
+```
+
+Note that text immediately below a table is treated as part of the table and
+will trigger this rule:
+
+```markdown
+| Heading | Heading |
+|---------|---------|
+| cell    | cell    |
```
DavidAnson, line 27: Nitpick: "Cell" for consistency

lib/md055.js
```diff
@@ -0,0 +1,26 @@
...
+const { filterByTypes } = require("../helpers/micromark.cjs");
+
+module.exports = {
+    "names": [ "MD055", "table-missing-border" ],
+    "description": "Table is missing leading or trailing pipe character",
+    "tags": [ "table" ],
+    "function": function MD055(params, onError) {
+        const tables = filterByTypes(params.parsers.micromark.tokens, ["table"]);
+        for (const table of tables) {
+            const rows = filterByTypes(table.children.flatMap((t) => t.children), ["tableRow", "tableDelimiterRow"]);
```
DavidAnson, line 15: No need for flatMap - filterByTypes looks at all children. Just pass table.

lib/md055.js
```diff
@@ -0,0 +1,26 @@
...
+            for (const row of rows) {
+                if (!row.text.startsWith("|")) {
+                    addErrorContext(onError, row.startLine, row.text, true);
+                }
+                if (!row.text.endsWith("|")) {
+                    addErrorContext(onError, row.startLine, row.text, null, true);
```
DavidAnson, line 21: false instead of null, please.

lib/md055.js
```diff
@@ -0,0 +1,26 @@
...
+
+module.exports = {
+    "names": [ "MD055", "table-missing-border" ],
+    "description": "Table is missing leading or trailing pipe character",
+    "tags": [ "table" ],
+    "function": function MD055(params, onError) {
+        const tables = filterByTypes(params.parsers.micromark.tokens, ["table"]);
+        for (const table of tables) {
+            const rows = filterByTypes(table.children.flatMap((t) => t.children), ["tableRow", "tableDelimiterRow"]);
+            for (const row of rows) {
```
DavidAnson, line 16: Nitpick: The first line inside the for block could be: `const { startLine, text } = row;` and then shorten the 6 following uses.

test/markdownlint-test.js
DavidAnson: ... and by "merge", I meant "rebase".

DavidAnson (changes requested):
Barely any comments from me this time! If you address this set and get the CI tests passing, I think we're good here. Thank you!

DavidAnson (changes requested):
Barely any comments from me this time! If you address this set and get the CI tests passing, I think we're good here. Thank you!

lib/md055.js
```diff
@@ -0,0 +1,27 @@
...
+const { filterByTypes } = require("../helpers/micromark.cjs");
+
+module.exports = {
+    "names": [ "MD055", "table-missing-border" ],
+    "description": "Table is missing leading or trailing pipe character",
+    "tags": [ "table" ],
+    "function": function MD055(params, onError) {
+        const tables = filterByTypes(params.parsers.micromark.tokens, [ "table" ]);
+        for (const table of tables) {
+            const rows = filterByTypes([ table ], [ "tableRow", "tableDelimiterRow" ]);
```
DavidAnson, line 15: I know I said to use table, but I should have said to use table.children - we know the table object won't match and table.children is already an array.

DavidAnson (changes requested):
I left one tiny comment. I think this is pretty much ready. From what I can tell, remaining work is to rebase onto latest next and make the tiny change I just added. If you'd like, I can do the rest, though I'll have to do it manually. I don't mind. Otherwise, in order to accept the PR we need to get everything happy here.

Actually, let me see if I have push permission to this branch - I may be able to get tests passing easily...

DavidAnson:
I don't want to rebase your branch without your permission, but I think that's all that's left before I do a final pass and complete the PR. Please let me know if that's okay to do.

gabriel-check24:
I'm ok with that, go ahead

gabriel-check24:
Yes, please, feel free to go ahead.

DavidAnson:
I’ve merged this PR into a new `md055` branch of the primary repo. I’m not merging it into `next` because I want to resolve the following issues:

1. GFM says leading/trailing pipe characters are optional - is it right for this rule to force them on? Is detection and a “consistent” style needed?
2. Related, there are far too many violations in the test repos - as a general rule I prefer not to introduce hundreds of violations unless the relevant Markdown is undeniably wrong.
3. A very brief review of test repo suggests a bug where spaces after the trailing pipe may cause a violation to be reported by this rule.
4. Is it practical to automatically fix violations?

You can PR against the `md055` branch if you wish, but it will be easier for me to iterate there than here.

Thank you for the contribution!

PR #1100, opened 2024-01-12
Support docfx tabs in MD025

Youssef1313:


DavidAnson:
I think the RegExp approach will be too fragile. The proposed version doesn't handle extra white space and links are especially tricky to handle in a regular expression. Instead, I can use the associated issue as an excuse to re-implement this rule using micromark and include the new functionality as part of that.

Youssef1313:
That would be very great! Thanks @DavidAnson 
Let me close this PR then.

PR #1109, opened 2024-01-25
Fix MD025 rule

oliviertassinari:
This rule was doing nothing on our codebase. Spotted this in https://github.com/mui/material-ui/pull/40784.

@samuelsycamore is you see recurring issues done on the docs or blog, feel free to invest time in markdown lint rules, this will help us scale quality content, we have a couple of custom ones on the codebase already (e.g. git diff format, straight quotes, table alignment, terminal language).


test/heading_mutliple_h1_no_toplevel.md
```diff
@@ -2,7 +2,7 @@ Some introductory text
 
 # Heading 1
 
-# Heading 2
+# Heading 2 {MD025}
 
 <!-- markdownlint-configure-file {
   "first-line-heading": false
```
oliviertassinari, line 8: The new regression test

lib/md025.js
```diff
@@ -23,7 +23,7 @@ module.exports = {
         if (hasTopLevelHeading || foundFrontMatterTitle) {
           addErrorContext(onError, token.lineNumber,
             token.line.trim());
-        } else if (token.lineNumber === 1) {
+        } else {
```
oliviertassinari, line 26: This condition made no sense to me.

DavidAnson:
Per the documentation, MD025 only applies when there is a top-level heading in the document, which it defines as an H1 on the first line:
https://github.com/DavidAnson/markdownlint/blob/main/doc/md025.md

The proposed edit in this PR seems to change that behavior to prevent multiple H1 regardless of where the first is located. I can see why one might want to enforce that, but it represents a change to how this rule has always behaved and will introduce new violations like the ones in this PR.

I do not think it is appropriate to redefine a rule after it has been in use for a long time, so I am inclined not to accept this change.

oliviertassinari:
> but it represents a change to how this rule has always behaved and will introduce new violations like the ones in this PR

I don't understand how the position of the h1 in the document is relevant to this role. This looks like a bug fix to me

Screenshot 2024-01-28 at 23 54 10

But fair enough, I need to create a custom rule anyway https://github.com/mui/material-ui/pull/40835. I'm closing. 👍 to add a params for such a case.

PR #1117, opened 2024-01-31
feat: support line number in fragment in MD051/link-fragments (e.g: '#L7')

theoludwig:
Hello! :smile:
I'm the original author of the rule MD051, work started in this PR: https://github.com/DavidAnson/markdownlint/pull/495.
I'm also the author of https://github.com/theoludwig/markdownlint-rule-relative-links custom rule.
This custom rule, verify if relative links exists in the file system using Node.js `node:fs`, it also checks for links fragments referencing other files (similar to the rule MD051, by trying to have as much as possible the same behavior as the MD051 built-in rule in this regard).

GitHub supports links fragments for referring to a specific line, e.g:
```md
[Link](#L7)
```
`#L7` means "Go To Line 7`.

More context: https://github.com/theoludwig/markdownlint-rule-relative-links/issues/6

Currently MD051 reports an error for that kind of fragments, but it should not, as they are valid on GitHub.

This PR fixes the issue, similarly to how it is done in `markdownlint-rule-relative-links`.
In the custom rule, we can go even further and check if the file has enough lines to be a valid line number reference (e.g: `#L7` in a empty file, is invalid, as line 7 doesn't exist), but that needs access to Node.js `node:fs`, which as far as I understand, you don't want to use for built-in rule for compatibility with the browser. So for now we're just considering it's valid without extra checks.

Feel free to suggest ideas, suggest better implementation, or maybe you have ideas of more test cases to add?

Thanks!


lib/md051.js
```diff
@@ -64,6 +64,34 @@ function unescapeStringTokenText(token) {
...
+ * Using `Number.parseInt` combined with `Number.isNaN` will not be sufficient enough because `Number.parseInt("1abc", 10)` will return `1` (a valid number) instead of `NaN`.
+ *
+ * @param {string} value String to check.
+ * @returns {boolean} `true` if string is a valid integer, `false` otherwise.
+ * @example isValidIntegerString("1") // true
+ * @example isValidIntegerString("45") // true
+ * @example isValidIntegerString("1abc") // false
+ * @example isValidIntegerString("1.0") // false
+ */
+function isValidIntegerString (value) {
```
theoludwig, line 90: I don't know if this function should be there?
Would it be good to move it elsewhere and do some proper unit testing specifically for this function only?

lib/md051.js
```diff
@@ -119,12 +147,27 @@ module.exports = {
...
             (text.length > 1) &&
             text.startsWith("#") &&
-            !fragments.has(text) &&
-            !fragments.has(`#${encodeURIComponent(text.slice(1))}`)
+            !fragments.has(textFragment)
           ) {
+            if (textFragment.startsWith("#L")) {
+              const textFragmentLowercase = textFragment.toLowerCase()
+              if (fragments.has(textFragmentLowercase)) {
+                // eslint-disable-next-line no-continue
```
theoludwig, line 159: I guess using `continue` in that context is fine? Nesting if blocks, makes the code difficult to read, and also generate a bigger diff, which is then harder to review.

theoludwig:
Test Repos seems failing by unrelated changes. As the errors are:
```sh
test-repos/v8-v8-dev/src/features/import-attributes.md: 59: MD034/no-bare-urls Bare URL used [Context: "https://chromestatus.com/featu..."]␊
  - test-repos/v8-v8-dev/src/features/import-attributes.md: 61: MD034/no-bare-urls Bare URL used [Context: "https://developer.apple.com/do..."]␊
  - test-repos/v8-v8-dev/src/features/import-attributes.md: 63: MD034/no-bare-urls Bare URL used [Context: "https://babeljs.io/blog/2023/0..."]␊
  ```
  
  Which MD034 is not updated with this PR.
  I've updated the snaptshots in this commit: [`d19f30a` (#1117)](https://github.com/DavidAnson/markdownlint/pull/1117/commits/d19f30a6fab914a5acff278d45f5221dd32f9d0e)

**Comment by DavidAnson** [comment]
Cool, thanks, I'll start a review in a moment!

A couple of things:
1. Is the GitHub behavior documented anywhere? It may be nice to mention this scenario in the docs for this rule and link to that page.
2. Thanks for respecting the desire to work on the web and not use node:fs. :)

**Comment by DavidAnson** [comment]
3. I fixed the test-repos issue. That specific test is sensitive because it pulls the latest from external repos and tests against them, failing if any violations have changed. This is deliberate on my part, so random failures in that test are a feature, not a bug.

File: lib/md051.js [hunk 3]
```diff
@@ -119,12 +147,27 @@ module.exports = {
...
           const text = unescapeStringTokenText(definition);
+          const textFragment = `#${encodeURIComponent(text.slice(1))}`
           if (
             (text.length > 1) &&
             text.startsWith("#") &&
-            !fragments.has(text) &&
-            !fragments.has(`#${encodeURIComponent(text.slice(1))}`)
+            !fragments.has(textFragment)
           ) {
+            if (textFragment.startsWith("#L")) {
```
DavidAnson, line 156: This `if` and everything in it feels like it could be replaced with something like:

```
const lineFragmentRe = /^#L\d+$/;
```

And then a single addition to the `if` checks above:

```
!lineFragmentRe.test(textFragment)
```

And remove the two functions added higher up in this file.

test/link-fragments.md
```diff
@@ -86,6 +86,12 @@
 
 [Valid][escapedref]
 
+[Valid](#L7)
```
DavidAnson, line 89: Can you please add another test case for multiple digits like "L#1234"?

test/link-fragments.md
```diff
@@ -86,6 +86,12 @@
...
+
+[Valid](#L12-abc)
+
+### L12 ABC
```
DavidAnson, line 93: Nitpick, but the headings are in roughly the same order as the links, so this belongs near the end of the valid heading list.

test/link-fragments.md
```diff
@@ -280,6 +286,8 @@ Valid Setext Heading with Named Fragment {#setext}
 
 [Invalid](#uppercase) {MD051}
 
+[Invalid](#L7abc) {MD051}
```
DavidAnson, line 289: Can you add another invalid case for "#L"?

test/snapshots/markdownlint-test-repos-small.js.md
```diff
@@ -321,6 +321,9 @@ Generated by [AVA](https://avajs.dev).
     test-repos/v8-v8-dev/src/features/globalthis.md: 43: MD033/no-inline-html Inline HTML [Element: feature-support]␊
     test-repos/v8-v8-dev/src/features/import-assertions.md: 84: MD034/no-bare-urls Bare URL used [Context: "https://chromestatus.com/featu..."]␊
     test-repos/v8-v8-dev/src/features/import-assertions.md: 88: MD034/no-bare-urls Bare URL used [Context: "https://github.com/babel/babel..."]␊
+    test-repos/v8-v8-dev/src/features/import-attributes.md: 59: MD034/no-bare-urls Bare URL used [Context: "https://chromestatus.com/featu..."]␊
```
DavidAnson, line 324: These go away after you rebase to latest.

DavidAnson (changes requested):
Makes sense to me! :) Please let me know if the simplification I propose works - I think it cuts out a lot of this code. Also, please mention the rule allows "#L123" because of GitHub in the docs. `doc-build/md051.md` is the file to edit and everything else will get generated by `npm run ci`.

lib/md051.js
theoludwig: This doesn't work, as `!lineFragmentRe.test(textFragment)` returns false with `#L12-abc`, but it could be a valid heading fragment (there is a test covering that).

However, we can still use this Regex, and replace the "2 functions" (`getLineNumberStringFromFragment` and `isValidIntegerString`), but we need to make sure that we ignore all `encodedText` matching `/^#L\d+$/` before verifying in `fragments` Map (still with a `continue`, I don't know if there's another easy alternative to do it), and then, we need to lower case `encodedText` to check if it is in `fragments` as the Map contains `#l12-abc`, not `#L12-abc`.

Talking about lowercase, I don't understand that test case:
```md
## Valid Fragments

[Title](#Valid-Fragments) {MD051}
```

`#Valid-Fragments` is valid, as GitHub doesn't care about case sensitivity for Heading Fragments, same validity as  `[Valid](#valid-fragments)`, so in this case the rule should not returns an error.
Is it for consistency that, the rule throws an error here?

Relaxing that part, could avoid that conditon:
```js
if (encodedText.startsWith('#L')) {
  encodedText = encodedText.toLowerCase();
}
```

And, instead of `inlineText.toLowerCase()` in `convertHeadingToHTMLFragment`, we could `.toLowerCase()` the whole return value.

lib/md051.js
```diff
@@ -119,12 +147,27 @@ module.exports = {
         for (const definition of definitions) {
           const { endColumn, startColumn } = definition;
           const text = unescapeStringTokenText(definition);
+          const textFragment = `#${encodeURIComponent(text.slice(1))}`
           if (
             (text.length > 1) &&
             text.startsWith("#") &&
-            !fragments.has(text) &&
```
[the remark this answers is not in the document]
theoludwig: The first thing, I've done, when investigating the code, is to remove this line, as I didn't understand why we need to check twice if `fragments.has` with both `text` and `encodedText`. I ran the tests; and everything works, and given the amount of tests cases, already covering this rule, it should pretty mean this line is pretty much "useless", and we don't need it, but I haven't look deep into it, to why it is the case.

theoludwig:
> Is the GitHub behavior documented anywhere? It may be nice to mention this scenario in the docs for this rule and link to that page.

Yes. It is documented here: https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-a-permanent-link-to-a-code-snippet#linking-to-markdown
I've added the explanation in `md051.md`. :smile:

lib/md051.js
DavidAnson: While there are multiple forms that work in practice, GitHub produces links for headings that are all lower-case, so that's what this rule wants to see.

DavidAnson:
The GitHub "specification" is pretty light on detail, but it was enough to remind me this syntax is more capable than just a single line number. For example, the following represents the full form I'm aware of which links to starting line/column and ending line/column: `https://github.com/DavidAnson/markdownlint/blob/9c77f92616c063a5d42f165cae2edb9e3418ac53/CHANGELOG.md?plain=1#L30C11-L31C9`

So the regular expression should be something more like `^L\d+(?:C\d+)?(?:-L\d+(?:C\d+)?)?$`. I'll look at the rest of the changes now, but this probably means we should update the code, tests, and docs.

doc-build/md051.md
```diff
@@ -38,6 +38,14 @@ attribute can be used to define a fragment:
 An `a` tag can be useful in scenarios where a heading is not appropriate or for
 control over the text of the fragment identifier.
 
+GitHub also supports links to [specific lines][github-lines-links].
+
+For example, to link to line 12 of current Markdown file:
```
DavidAnson, line 43: I'm not sure I want to suggest doing this because GitHub also says you need to do `plain=1`. I'd say let's say something like "That syntax looks like this is practice:".

doc-build/md051.md
```diff
@@ -38,6 +38,14 @@ attribute can be used to define a fragment:
 An `a` tag can be useful in scenarios where a heading is not appropriate or for
 control over the text of the fragment identifier.
 
+GitHub also supports links to [specific lines][github-lines-links].
```
DavidAnson, line 41: At the end: "... so this rule will not report violations when this syntax is used.".

lib/md051.js
```diff
@@ -119,11 +119,22 @@ module.exports = {
...
           const text = unescapeStringTokenText(definition);
+          let encodedText = `#${encodeURIComponent(text.slice(1))}`;
+
+          const lineFragmentRe = /^#L\d+$/;
+          if (lineFragmentRe.test(encodedText)) {
+            // eslint-disable-next-line no-continue
+            continue;
+          }
+
+          if (encodedText.startsWith('#L')) {
```
DavidAnson, line 130: I don't understand this part. Why does this need special handling? I'd assume if this were not done then we could move the lineFragmentsRe.test into the if below along with the other checks as I originally proposed. There was no lowercase conversion before, so I think there should not be one after. Maybe this gets to the other point about case sensitivity - I consider the "L111" form to be the only allowed form by this rule since that's what GitHub produces. Anyone using "l111" should switch - which will help avoid confusing letters and numbers. :)

DavidAnson (changes requested):
I like how this change shrunk, thank you! The docs are good to have as well. Biggest item in this review pass is the full LC-LC GitHub syntax from my separate comment.

lib/md051.js
theoludwig: Without this condition, the following test case fails:
```md
### L12 ABC

[Valid](#L12-abc)
```

From what I understand, with your previous lowercase related comment, it is to be expected, as the link should have been `[Valid](#l12-abc)` as it is a heading.
But when it's about `#L12`, the L should be uppercase.

Looks fine to me, let's be stricter than GitHub about case sensitivity. But we should maybe document it in `md051.md`? As it might be unexpected for users (me included) that the rule reports errors for valid fragments because of stricter case sensitivity management.

Will simplify the code and make the necessary changes whenever I have time. :+1:

theoludwig:
@DavidAnson Thank you for reviewing the PR. :smile:
Sorry for the delay.

I should have addressed most feedbacks from reviews, notably:
- Updated the documentation, to specify that the rule mandates the use of lowercase for link fragments for headings.
- Simplified the code, as we don't support anymore uppercase/lowercase, e.g: `[Link](#L7abc)`, it should be `[Link](#l7abc)` with a valid heading.
- Support specifying columns (e.g: `[Valid](#L30C11-L31C9)`).

DavidAnson:
This looks good, thank you! I think I want to iterate on the documentation a little after thinking about it some more - I'll do that tomorrow by pushing to this branch and then accept the PR.

DavidAnson:
For posterity since I tried the combinations out, these are all the possibilities and how GitHub handles them:

HIGHLIGHT
#L30C11-L31C9
#L30C11-L31
#L30-L31
#L30-L31C9
#L30

NO HIGHLIGHT
#L30C11-C9
#C11-L31C9
#L30C11

IGNORED
#C11
#C11-C9

DavidAnson:
... and this regular expression matches ONLY the top group above: `^(?:#L\d+(?:C\d+)?-L\d+(?:C\d+)?)|(?:#L\d+)$`

DavidAnson (approved):
Applied the tweaks and will accept this PR in a moment - thank you again!!

PR #1160, opened 2024-03-24
fix: Export types

regseb:
I'm using Markdownlint in my project and TypeScript is reporting the following problem:

```
node_modules/markdownlint/lib/markdownlint.d.ts:232:34 - error TS7016: Could not find a declaration file for module 'markdownlint-micromark'. '/home/regseb/dev/metalint/node_modules/markdownlint-micromark/micromark.cjs' implicitly has an 'any' type.
  Try `npm i --save-dev @types/markdownlint-micromark` if it exists or add a new declaration (.d.ts) file containing `declare module 'markdownlint-micromark';`

232 type MicromarkTokenType = import("markdownlint-micromark").TokenType;
```

I think [`exports`](https://nodejs.org/docs/latest/api/packages.html#exports) and [`types`](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html) are _incompatible_. The pairs are :

- `main` / `types`
- `exports.default` / `exports.types`

### DIFF
--- a/micromark/package.json
+++ b/micromark/package.json
@@ -3,7 +3,13 @@
   "version": "0.1.9",
   "description": "A trivial package that re-exports some micromark functionality as a CommonJS module",
   "type": "commonjs",
-  "exports": "./micromark.cjs",
+  "exports": {
+    ".": {
+      "types": "./micromark.d.ts",
+      "default": "./micromark.cjs"
+    }
+  },
+  "main": "./micromark.cjs",
   "types": "./micromark.d.ts",
   "author": "David Anson (https://dlaa.me/)",
   "license": "MIT",


DavidAnson:
Interesting, thanks! This doesn't seem to happen within the repository. How are you reproducing it? I'd like to do something similar to make sure the fix is good and this doesn't regress. I use https://github.com/DavidAnson/markdownlint/blob/main/example/typescript/type-check.ts and it has helped catch problems like this one in the past.

DavidAnson:
And to confirm, this is only a warning from the compiler and not an error, right?

regseb:
This is an error that is only reported with [`strict`](https://www.typescriptlang.org/tsconfig#strict) flag.

---

To reproduce:

- `package.json`

  ```JSON
  {
    "name": "testcase",
    "version": "1.0.0",
    "type": "module",
    "dependencies": {
      "markdownlint": "0.34.0",
      "typescript": "5.4.3"
    }
  }
  ```

- `index.ts`

  ```TypeScript
  import markdownlint from "markdownlint";

  console.log(markdownlint);
  ```

1. `npm install`
2. `npx tsc --moduleResolution nodenext --module nodenext --strict index.ts`

DavidAnson:
I can reproduce this in the repo with `type-check.ts` which already uses `strict`, but **only** after changing `module` and `moduleResolution` to `NodeNext` as passed on the command-line in your example: <https://github.com/DavidAnson/markdownlint/blob/main/example/typescript/tsconfig.json>

I need to read the documentation you link to and update the use of `type-check.ts` to cover both scenarios before I will be ready to accept this PR.

DavidAnson:
This is interesting. TypeScript seems to ONLY document "types" (your second link) while Node.js documents "exports.types" but only as a community condition definition they explicitly do not take responsibility for. I don't see any documentation that calls out the incompatible pairing you talk about. But I see various TypeScript issues complaining about very related issues. :)

DavidAnson:
Closing this PR as the linked commits above address the reported issue in a different way. Sorry, I couldn't easily reuse this PR because the process of testing and published across the two packages was somewhat complicated. Thanks for raising this!

PR #1299, opened 2024-07-17
MD018: do not check `#` in HTML blocks

### LINKED ISSUE #1268: MD018 error for ID selectors in inline CSS
```markdown
# Test

<div id="my-table">
<style>
#my-table table {
    width: 100%;
}
</style>

|                                  | A          |    B             |
| -------------------------------- | :--------: | :--------------: |
|                                  |  C         |   D              |

</div>
```

This generates the error:

```
markdownlint-cli2 v0.13.0 (markdownlint v0.34.0)
Finding: a.md
Linting: 1 file(s)
Summary: 1 error(s)
a.md:5:1 MD018/no-missing-space-atx No space after hash on atx style heading [Context: "#my-table table {"]
```

fyc09:
Fix #1268.

Side Effect: MD013 will treat html code (including embedded markdownlint-configure-file) as code blocks, which will emit errors in these cases:

https://github.com/DavidAnson/markdownlint/blob/b2305efafb034b1f328845aec9928b5363ffd646/test/long-lines-short-code.md?plain=1#L27-L32
https://github.com/DavidAnson/markdownlint/blob/b2305efafb034b1f328845aec9928b5363ffd646/test/long-lines-short-code.md?plain=1#L27-L32
https://github.com/DavidAnson/markdownlint/blob/b2305efafb034b1f328845aec9928b5363ffd646/test/long-lines-strict.md?plain=1#L33-L39


DavidAnson:
"Code" in the case of getLineMetadata is meant to refer to one of the two Markdown code block types (indented or fenced). The proposed change could have more side effects beyond what you have already identified. I had been thinking that a fix specific to MD018 would be best here. Did you try that first?

DavidAnson:
By the way, could you please update the test file to include scenarios that would trigger rules 18, 19, 20, and 21? If 18 is broken, then I suspect 20 is also. 19 and 21 are already migrated to micromark in the next branch, so should be OK as-is. But I would like the new test file to verify the HTML scenario for all four of these similar rules. Thank you!

fyc09:
I'm unsure if we should investigate this violation, as the current implementation will simply ignore it.

```markdown
##Title <span>text</span> {MD018} {MD033}
```

DavidAnson:
What you show above seems like correct behavior, or am I missing something?

test/ignore-html-block.md
```diff
@@ -0,0 +1,15 @@
+# ignore-html-block.md
+
+## Title
+
+##Title2 {MD018}
```
DavidAnson, line 5: Let's please remove all the "Title" lines so everything is testing HTML blocks.

test/ignore-html-block.md
```diff
@@ -0,0 +1,15 @@
...
+
+<div> {MD033}
+#foo
```
DavidAnson, line 8: "#heading", please. I try not to not use "foo"/etc.

test/ignore-html-block.md
```diff
@@ -0,0 +1,15 @@
...
+</div>
+
+##Title3 {MD020}##
+
+<div> {MD033}
+#foo#
+</div>
```
DavidAnson, line 15: Could you please add new blocks for the MD019/MD021 scenarios?

DavidAnson (changes requested):
Thank you, this looks good! I left some comments on the generated file which was a mistake. Please make the suggested changes to the source files instead.

DavidAnson:
FYI, there will be a failure in TeatRepos because that is failing in next branch. I will fix that shortly, so please ignore for now.

fyc09:
> What you show above seems like correct behavior, or am I missing something?

Currently, only MD033 is emitted; MD018 is ignored due to the HTML block in this line. I think emitting both MD033 and MD018 will be more reasonable.

DavidAnson:
I agree. But the above is an HTML span, not a block. I would not expect it to be affected by the new code?

fyc09:
> I agree. But the above is an HTML span, not a block. I would not expect it to be affected by the new code?

You are right, I made an mistake.

DavidAnson:
Please rebase PR to latest next branch commit to fix "sh: 1: ava: not found". This is a bug in latest Node/npm which I have worked around.

DavidAnson:
I can't easily view the snapshot diff on my phone, but all the other code changes look good, thank you!

DavidAnson:
I had a look at the differences, and there is way more going on than there needs to be. After you re-base (see comment above), please run "npm run update-snapshots" to regenerate that file. The diff at that point should be ONLY the errors and corrections for the new test file you added. After that, we make sure checks pass and I'll approve. :)

fyc09:
I reran "npm run update-snapshots", but the snapshot didn't change, and I enabled GitHub action in my own repository and the [CI](https://github.com/fyc09/markdownlint/actions/runs/9993147684) looks good. I think the tests' order is changed, but I do not know what to do.

DavidAnson:
I wonder if this is a locale issue that sorts differently for you? I will give it a try in a day or two and either push a commit to your branch with the "right" order or approve your changes as-is. Thanks again!

DavidAnson (approved):
Updating snapshots on my end worked.

PR #1315, opened 2024-07-31
Group minor/patch Dependabot updates

nschonni:
Major versions will still create individual PRs, but minor/patch ones are grouped together like https://github.com/canada-ca/ore-ero/pull/1915

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,9 +11,23 @@ updates:
       interval: "daily"
     target-branch: "next"
     versioning-strategy: "increase"
+    groups:
+      dependencies:
+        patterns:
+          - "*"
+        update-types:
+          - "minor"
+          - "patch"
   - package-ecosystem: "npm"
     directory: "/micromark"
     schedule:
       interval: "daily"
     target-branch: "next"
     versioning-strategy: "increase"
+    groups:
+      micromark:
+        patterns:
+          - "*"
+        update-types:
+          - "minor"
+          - "patch"


DavidAnson:
Thank you! I'm glad this capability exists, but I actually do appreciate that the current approach gives each dependency change its own PR and its own validation once accepted. It's a bit more work for me to accept things individually, but I'm OK with it and that's how I'm inclined to stay for now.

nschonni:
That's fine, the other part of this, is that even though CI is free, this reduces the load and rebase churn

DavidAnson:
It's a good point, but GitHub has never complained so far and I don't know how to tell Dependabot to split a PR if something fails. :)

Something that feels wasteful for no reason is how the checks run twice sometimes because a) branch was pushed and b) PR was updated. I want each of these events to cause checks to run, but don't need duplicate checks when the same commit happens to tick both boxes. I haven't looked into this, but suspect there is a way to be more efficient.

nschonni:
> It's a good point, but GitHub has never complained so far and I don't know how to tell Dependabot to split a PR if something fails. :)

Seems kind of like the current way of giving it a mention:

> `@dependabot ignore <dependency name> minor version` will close this group update PR and stop Dependabot creating any more for the specific dependency's minor version (unless you unignore this specific dependency's minor version or upgrade to it yourself)

> Something that feels wasteful for no reason is how the checks run twice sometimes because a) branch was pushed and b) PR was updated. I want each of these events to cause checks to run, but don't need duplicate checks when the same commit happens to tick both boxes. I haven't looked into this, but suspect there is a way to be more efficient.

I think you're OK for that case, as Dependabot Pushes aren't built https://github.com/DavidAnson/markdownlint/blob/b2305efafb034b1f328845aec9928b5363ffd646/.github/workflows/ci.yml

Feel free to close this, but I'll keep it open for now

Skandiakiwi (commented):
Looks good 😊

Skandiakiwi (commented):
Good to go

DavidAnson:
I'm closing this because I don't see myself deciding to merge these unless something else changes. Thanks for the suggestion!

PR #1341, opened 2024-08-26
Fix typo update README.md

bravesasha:
Fix typo update README


DavidAnson:
The proposed change would break the link. Also, the project is called "webhint".

PR #1389, opened 2024-10-22
chore: Create GitHub Issue templates

nschonni:
Had this sitting around in drafts for awhile, but I'm not sure if it needs more work before you might accept it.
- Adds the Issue config, so users will have a little nudge to open Issues on the "correct" repo, instead of this one to start
- Use the new YAML forms to allow a little guided data collection
- Still allows the "blank" issue templates, if people don't like the config

I'm opening this on `main` instead of `dev` as the templates need to be on the default branch to work. Could open it against `dev`, but then they wouldn't show up till the next sync

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -1,0 +1,37 @@
+name: Bug Report
+description: File a bug report.
+title: "[Bug]: "
+labels: ["bug", "triage"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thanks for taking the time to fill out this bug report!
+  - type: textarea
+    id: what-happened
+    attributes:
+      label: What happened?
+      description: Also tell us, what did you expect to happen?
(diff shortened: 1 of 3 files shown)


DavidAnson:
Interesting ideas here, let me think about this! (Most of my own experience with templates has been more negative than positive.) (I try to never touch `main` branch, so I would want this in `next` as you suggest.)

DavidAnson:
I think I'll merge these ideas into a very simple template based on the original format. (Forms is apparently not official: <https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms>) There doesn't seem to be a way to preview issue templates, so I'll iterate in a different repo, then credit you in the commit.

nschonni:
No need for credit :)
You can also use the builder in the repo Settings, and that will create the older Markdown based templates for you. You could pull in the `config.yml` separately if you like the generic links/descriptions form that file

PR #1396, opened 2024-10-31
ci: Add Node 23

nschonni:
Think it was mentioned on another thread that ava might have an issue


DavidAnson:
I used to test the "in development" version (odd version number) of Node, but I stopped because I don't really want to spend time debugging temporary problems that aren't mine. The pre-commit failures you reference seem to be due to a path truncation bug and whatever is going on here seems to be some kind of incompatibility with TypeScript types. While this issue may be fixable with a JSDoc change in project code, I doubt there's much I can do about the path issue.

DavidAnson:
OK, wait, the failure in this run seems to have hit the nightly main branch CI run as well, so it's probably unrelated to Node 23. That said, my comments above about staying off of the development branches of Node remain true.

nschonni:
Yeah, I've been around Node for a long time too and don't personally use the odd releases either. I think the concept of them being a dev release went away awhile, abd now it's just a regular release that doesn't become LTS

DavidAnson:
It looks like you're right about that policy change and I am happy to accept this pull request. However, GitHub is saying it's not merge-able in its current state for some reason and I wonder if you could maybe rebase it or something to make it happy?

nschonni:
I wasn't seeing the conflict message in the UI, but I gave it rebase against `next` to see if that solves it on your side

DavidAnson:
Thank you! It’s apparently just in the app and it wanted me to merge as administrator for some reason.

nschonni:
Maybe just a caching thing. The original push had a failing build and got marked as not meeting the branch policy, so was asking for an admin override. That's just a guess though

Issue #681, opened 2022-12-15
[Proposal] Rule to discourage generic link text


khiga8:
Hello! First off, thank you for this project! 

I wanted to suggest an accessibility-focused rule that discourages generic link text from being set such as `Read more`, `Learn more`, `link`, `Click here`, `here`, etc. 

Screen reader users often navigate through a page by tabbing through links, or by using a screen reader menu that displays all the links on a page. If the link text is super generic like, `Learn more`, it's not clear where that link will go to without needing to navigate the entire surrounding context. Descriptive link texts are beneficial even for non-screen reader users.

[Access Guide: Descriptive link text](https://www.accessguide.io/guide/descriptive-link-text)

I'm thinking that we could have a default array of discouraged link text, and allow additional ones to be configured.

I'm happy to take this work on. Let me know what you think!

DavidAnson:
This makes sense and should be pretty straightforward since it can use the parser output for links directly. There is precedent in existing rules for analyzing link content that could inform the implementation of this rule. There is already an "accessibility" tag that can be used here. I plan to do a new release soon and will not include this due to timing, but you could iterate in a PR if you'd like. Please see CONTRIBUTING.md or similar PRs for expectations. Thanks!

Ahmatdarjan123:
Goddddddd

Ahmatdarjan123:
Goddddddddd

Ahmatdarjan123:
Godddddd

khiga8:
2 years late, but I'll pick this up :)

DavidAnson:
Haha, these days I recommend the micromark parser. The custom rule docs are up to date!

DavidAnson:
(I guess you are proposing to add this to the framework, so you'll HAVE to use that parser. :) Please review the contributing guidelines and share early/often so we can avoid misunderstandings.)

khiga8:
Sounds good!

I just opened [[WIP] Add new rule: no-generic-link-name #1459](https://github.com/DavidAnson/markdownlint/pull/1459), if you'd like to provide a first pass review to ensure I'm on the right track! :)

PR #1459, opened 2025-01-06
Add new rule: `no-generic-link-name`

khiga8:
Closes: https://github.com/DavidAnson/markdownlint/issues/681

This PR introduces a new lint rule to discourage generic, unhelpful link names. More context in linked issue!

While this rule has a default set of "banned names", consumers have the option to override the default and set their own banned words through the `link_texts` config if they'd like.


lib/md039.mjs
```diff
@@ -35,18 +36,6 @@ function addLabelSpaceError(onError, label, labelText, isStart) {
...
- * @param {Map<string, any>} definitions Map of link definitions.
- * @returns {boolean} True iff the link is valid.
- */
-function validLink(label, labelText, definitions) {
-  return (label.parent?.children.length !== 1) || definitions.has(labelText.text.trim());
-}
-
 /** @type {import("markdownlint").Rule} */
 export default {
   "names": [ "MD039", "no-space-in-links" ],
```
khiga8, line 41: I closely copied this rule's logic since it looked similar to what we're trying to (aka parse the link text).
I ended up moving this `validLink` to a shared helper.

khiga8:
I'm not sure how to interpret this failure:

```
 test › markdownlint-test › validateJsonUsingConfigSchemaStrict

  Rejected promise returned by test. Reason:

  SyntaxError {
    message: 'Unexpected token } in JSON at position 42',
  }
```

I'd appreciate pointers!

DavidAnson:
Great! I'll make a first pass review later tonight. Regarding the test failure you ask about, I think it is due to the trailing commas added to the new configuration comment at the bottom of test/long-lines-long-reference-definitions.md. JSON does not allow trailing commas.

In general, I think this would be the first built-in rule which only works in a single language (English) in its default configuration. I'm not sure how I feel about that... Certainly, all the documentation is in English because that's all I know and it's fairly common as a default language in the industry for documentation and code. However, to have the linting tool itself start to "favor" one language over another feels a little weird?

I'm open to counter-arguments or examples of another built-in rule that behaves like this. At the same time, having this rule default to an empty list would make it less useful. (Though there is precedent for that, see MD044/proper-names.) I do not think I want to start maintaining multiple translations for the phrase list because there are a lot of languages out there and I can only validate English for correctness.

DavidAnson:
I was curious how a similar-ish project handled this situation: https://github.com/get-alex/alex/issues/202

Ugh, that is not really a road I went to start down. My current thinking is to do something similar to MD044/proper-names and default this rule to an empty list.

I'm still open to feedback, though!

helpers/helpers.cjs
```diff
@@ -540,3 +540,18 @@ function expandTildePath(file, os) {
...
+/**
+ * Downcases a string and strips extra white spaces and punctuations
+ *
+ * @param {string} text String to transform
+ * @returns {string} Stripped and downcased string
+ */
+function stripAndDowncaseText(text) {
+  return text
+    .toLowerCase()
+    .replace(/[.,/#!$%^&*;:{}=\-_`~()]/g, "")
```
DavidAnson, line 553: Can't this be `\W` (not word) or else `\p{whatever}` for character class? For example, why aren't square brackets included here?

helpers/helpers.cjs
```diff
@@ -540,3 +540,18 @@ function expandTildePath(file, os) {
...
 }
 module.exports.expandTildePath = expandTildePath;
+
+/**
+ * Downcases a string and strips extra white spaces and punctuations
+ *
+ * @param {string} text String to transform
+ * @returns {string} Stripped and downcased string
+ */
+function stripAndDowncaseText(text) {
```
DavidAnson, line 550: Once this function is moved, I think it can be renamed something like "normalize". Otherwise, I'll want to remove the word "strip" for possible connotations.

helpers/helpers.cjs
```diff
@@ -540,3 +540,18 @@ function expandTildePath(file, os) {
   return homedir ? file.replace(/^~($|\/|\\)/, `${homedir}$1`) : file;
 }
 module.exports.expandTildePath = expandTildePath;
+
+/**
+ * Downcases a string and strips extra white spaces and punctuations
```
DavidAnson, line 545: Period at the end of all these, please, even though I know they're not really sentences. Holdover from .NET days...

helpers/helpers.cjs
```diff
@@ -540,3 +540,18 @@ function expandTildePath(file, os) {
...
+ * @returns {string} Stripped and downcased string
+ */
+function stripAndDowncaseText(text) {
+  return text
+    .toLowerCase()
+    .replace(/[.,/#!$%^&*;:{}=\-_`~()]/g, "")
+    .replace(/\s+/g, " ")
+    .trim();
+}
+module.exports.stripAndDowncaseText = stripAndDowncaseText;
```
DavidAnson, line 557: Having converted so much code to ESM, `module.exports` gives me the ick now... :)

helpers/micromark-helpers.cjs
```diff
@@ -284,6 +284,18 @@ const nonContentTokens = new Set([
...
 
+/**
+ * Determines if a link is a valid link (and not a fake shortcut link due to parser tricks).
+ *
+ * @param {Token} label Label token.
+ * @param {Token} labelText LabelText token.
+ * @param {Map<string, any>} definitions Map of link definitions.
+ * @returns {boolean} True iff the link is valid.
+ */
+function validLink(label, labelText, definitions) {
```
DavidAnson, line 295: I think this function is no longer necessary after a recent refactoring. I'm about to remove it from the code, so will question whether it's necessary in your scenario, either.

lib/md059.mjs
```diff
@@ -0,0 +1,41 @@
+// @ts-check
+
+import { addErrorContext, stripAndDowncaseText } from "../helpers/helpers.cjs";
+import { validLink } from "../helpers/micromark-helpers.cjs";
+import { getReferenceLinkImageData, filterByTypesCached } from "./cache.mjs";
+
+const defaultBannedNames = [
```
DavidAnson, line 7: This should probably be a Set?

lib/md059.mjs
```diff
@@ -0,0 +1,41 @@
...
+  "read more",
```
DavidAnson, line 8: Maybe this goes away, but sort alphabetically if not, please.

lib/md059.mjs
```diff
@@ -0,0 +1,41 @@
...
+  "names": [ "MD059", "no-generic-link-name" ],
+  "description": "Links should have a descriptive name",
+  "tags": [ "links", "accessibility" ],
+  "parser": "micromark",
+  "function": function MD059(params, onError) {
+    const { definitions } = getReferenceLinkImageData();
+
+    // Allow consumer to define their own banned names list.
+
+    const bannedNames = params.config.banned_names_override || defaultBannedNames;
```
DavidAnson, line 27: I think calling it `banned_names` is even better.

lib/md059.mjs
```diff
@@ -0,0 +1,41 @@
...
+    const labels = filterByTypesCached([ "label" ])
+      .filter((label) => label.parent?.type === "link");
+
+    for (const label of labels) {
+      const labelTexts = label.children.filter((child) => child.type === "labelText");
+      for (const labelText of labelTexts) {
+        const text = labelText.text;
+        if (validLink(label, labelText, definitions) && bannedNames.includes(stripAndDowncaseText(text))) {
+          addErrorContext(onError, labelText.startLine, text);
```
DavidAnson, line 36: We can provide specific context and range information, too, I think - see what MD039 does.

lib/md059.mjs
```diff
@@ -0,0 +1,41 @@
...
+    // Allow consumer to define their own banned names list.
+
+    const bannedNames = params.config.banned_names_override || defaultBannedNames;
+    const labels = filterByTypesCached([ "label" ])
+      .filter((label) => label.parent?.type === "link");
+
+    for (const label of labels) {
+      const labelTexts = label.children.filter((child) => child.type === "labelText");
+      for (const labelText of labelTexts) {
+        const text = labelText.text;
```
DavidAnson, line 34: `const { text } = label;` for consistency.

test/link-style-no-url-inline-not-possible.md
```diff
@@ -60,6 +60,7 @@ Text [email] text
 [email]: user@example.com
 
 <!-- markdownlint-configure-file {
+  "no-generic-link-name": false,
```
DavidAnson, line 63: I think these are generally sorted alphabetically.

test/long-lines-long-reference-definitions-stern.md
```diff
@@ -1,5 +1,6 @@
 # Long Lines Long Reference Definitions Stern
 
+<!-- markdownlint-disable-file no-generic-link-name -->
```
DavidAnson, line 3: Bottom of the file, please, since `-file` applies to the entire file by nature.

test/long-lines-long-reference-definitions.md
```diff
@@ -15,3 +16,9 @@
 [long-reference-definition-image]: https://example.com/long/long/long/long/long/long/long/long/long/long/long/long/long/image
 <!-- markdownlint-disable-next-line link-image-reference-definitions -->
 [long-reference-definition]: https://example.com/long/long/long/long/long/long/long/long/long/long/long/long/long
+
+<!-- markdownlint-configure-file {
```
DavidAnson, line 20: Adding this shouldn't be necessary? Also, I think this is the source of the invalid JSON test failure.

test/markdown-in-html.md
```diff
@@ -34,4 +34,4 @@ Text
 
 </p>
 
-<!-- markdownlint-disable-file line-length no-inline-html -->
+<!-- markdownlint-disable-file line-length no-inline-html no-generic-link-name  -->
```
DavidAnson, line 37: Sort, please.

test/no-generic-link-name-config-override.md
```diff
@@ -0,0 +1,17 @@
+# Links with descriptive and non-descriptive names
```
DavidAnson, line 1: In newer tests, I try to make the top-level heading text match the file name, but in its readable form.

test/no-generic-link-name.md
```diff
@@ -0,0 +1,19 @@
+# Links with descriptive and non-descriptive names
+
+[Learn about Javascript](https://example.com/javascript/about)
```
DavidAnson, line 3: These are all inline links - please add some reference and shortcut links, too.

test/no-generic-link-name.md
```diff
@@ -0,0 +1,19 @@
...
+
+[Learn more](https://example.com/javascript/about). {MD059}
+
+[Click here](https://example.com/javascript/about). {MD059}
+
+To learn more, go [here!](https://example.com/javascript/about). {MD059}
+
+To learn more, [click here!!!!](https://example.com/javascript/about). {MD059}
+
+[click here!!!!](https://example.com/javascript/about). {MD059}
```
DavidAnson, line 17: How about some punctuation between the words of a banned phrase? "click-here"

test/table-content-with-issues.md
```diff
@@ -9,5 +9,5 @@
 | * emphasis*                  | {MD037} |
 | __strong __                  | {MD037} |
 |  ` code`                     | {MD038} |
-| [link ](https://example.com) | {MD039} |
-| [link]()                     | {MD042} |
+| [link ](https://example.com) | {MD039} {MD059} |
```
DavidAnson, line 12: Could you expand the table rows above so the columns are consistent? I like this and want to keep the new test case.

DavidAnson (changes requested):
Okay, it may look like a lot, but most of the changes are pretty small. :) Thanks for starting this out - let's continue the conversation around naming and default value in the body of the PR.

helpers/micromark-helpers.cjs
DavidAnson: It's gone and you should be able to simplify similarly: https://github.com/DavidAnson/markdownlint/commit/637975f517cfe856d034cc27951cf2afaf4186a3

README.md
```diff
@@ -146,6 +146,7 @@ playground for learning and exploring.
 - **[MD055](doc/md055.md)** *table-pipe-style* - Table pipe style
 - **[MD056](doc/md056.md)** *table-column-count* - Table column count
 - **[MD058](doc/md058.md)** *blanks-around-tables* - Tables should be surrounded by blank lines
+- **[MD059](doc/md059.md)** *no-generic-link-name* - Links should have a descriptive name
```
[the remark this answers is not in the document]
khiga8: Ooh, I like that positive phrasing :)

lib/md059.mjs
khiga8: Updating this to `banned_names` will give me an ESLint error - `Identifier 'banned_names' is not in camel case.`

schema/markdownlint-config-schema-strict.json
```diff
@@ -1716,6 +1716,42 @@
       "type": "boolean",
       "default": true
     },
+    "MD059": {
+      "description": "MD059/descriptive-link-text : Link text should be descriptive : https://github.com/DavidAnson/markdownlint/blob/v0.37.3/doc/md059.md",
```
khiga8, line 1720: Not sure what this should be set to, yet. Same for below.

lib/md059.mjs
khiga8: Opted to change  [Line 37](https://github.com/DavidAnson/markdownlint/pull/1459/files#diff-6324bd4331648f082986c51ad43c2fe1770b5f0461eb4347c7dfe25262e3107aR37) to a set!

README.md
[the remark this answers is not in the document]
khiga8: I went ahead and replaced `no-generic-link-name` with `descriptive-link-text`. Happy to update if we find a name we like better!

schema/markdownlint-config-schema-strict.json
```diff
@@ -1716,6 +1716,42 @@
...
+        "object"
+      ],
+      "properties": {
+        "banned_link_texts": {
+          "description": "List of banned link texts",
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "default": []
```
khiga8, line 1732: TBD

khiga8:
Hey @DavidAnson, thank you for your speedy and thorough review!

I like your suggestion of renaming this to `descriptive-link-text`, so I've made corresponding updates in this PR.

> Certainly, all the documentation is in English because that's all I know and it's fairly common as a default language in the industry for documentation and code. However, to have the linting tool itself start to "favor" one language over another feels a little weird?

> My current thinking is to do something similar to MD044/proper-names and default this rule to an empty list.

For sure, these are valid concerns! 

My _personal_ preference is for setting a default list for the English language (since we have a general sense of common, "non-descriptive names" for the following reasons:

* If consumers of this project want to expand on the default words, disagree with the defaults, or add support for additional languages, they have the option to do so through the config. 
* Setting some kind of default would make this rule useful for more projects immediately and help spread awareness around best a11y practices versus having to discover this through the docs.

I guess there's also a question around whether this rule should be on/off by default (not sure if you have a standard procedure for new rules?) If this rule is on by default with a list of defaults, some projects will encounter a lot of failure when bumping up the version. But then again, maybe that's okay because if a project doesn't feel ready to enforce the rule, they can turn it off. 

Let me know what you think!

lib/md059.mjs
khiga8: TODO

test/descriptive-link-text.md
```diff
@@ -0,0 +1,23 @@
...
+
+To learn more, go [here!](https://example.com/javascript/about). {MD059}
+
+To learn more, [click here!!!!](https://example.com/javascript/about). {MD059}
+
+[click-here!!!!](https://example.com/javascript/about). {MD059}
+
+Go to this [link]((https://example.com/javascript/about)). {MD059}
+
+[link][1] {MD059}
```
khiga8, line 21: @DavidAnson Is this what you meant by reference link? 
Do you have an example of a "shortcut link"?

related: https://github.com/DavidAnson/markdownlint/pull/1459#discussion_r1904887361

schema/markdownlint-config-schema-strict.json
khiga8: Whoops looks like I was not supposed to manually update it. I see a few potentially relevant scripts. Undoing!

lib/md059.mjs
khiga8: Updated in [02d1166](https://github.com/DavidAnson/markdownlint/pull/1459/commits/02d116629e5417d922046572307c6a6ee0414c6d)!

lib/md059.mjs
DavidAnson: I'm referring to the config parameter - instead of `banned_names_override`, just `banned_names`.

test/descriptive-link-text.md
DavidAnson: Examples here: https://github.com/DavidAnson/markdownlint/blob/main/doc/md054.md

DavidAnson:
All rules are enabled by default and I'd like to keep it that way. You make a good point about this being more discoverable if there are default values. Then I'd want to document that there will ONLY be English defaults so people don't PR Klingon or something. I don't mind it breaking for existing sites, but I want to be sure the default list is super obviously bad. Like "here" bad. People can disagree, but nobody should argue that "here" is a good link text. :)

lib/md059.mjs
```diff
@@ -0,0 +1,64 @@
...
+      .filter((label) => label.parent?.type === "link");
+
+    for (const label of labels) {
+      const labelTexts = label.children.filter((child) => child.type === "labelText");
+      for (const labelText of labelTexts) {
+        const { text } = label;
+        if (bannedNames.has(normalizeText(text))) {
+          const range = [
+            labelText.startColumn,
+            text.length
```
DavidAnson, line 49: BUT if `startLine` !== `endLine`, then this is a bad range. There should be examples elsewhere where I set `range` to `undefined`, then check the lines, then set `range` iff it's a single line. If you can't find examples, let me know. This should also be a test case - something like this will throw with the code shown here:
```
[link
text](blah)
```

test/bare-urls.md
```diff
@@ -12,7 +12,7 @@ Visit https://example.com, then refresh. {MD034}
 
 The site (https://example.com) is down. {MD034}
 
-<!-- markdownlint-disable line-length no-inline-html -->
+<!-- markdownlint-disable line-length no-inline-html descriptive-link-text -->
```
DavidAnson, line 15: Sort, please :)

test/descriptive-link-text-override-config.md
```diff
@@ -0,0 +1,17 @@
+# Descriptive link text override config
```
DavidAnson, line 1: I'm getting nitpicks now, but please remove "config" from the end - it's just an override. (Sorry!)

test/descriptive-link-text-override-config.md
```diff
@@ -0,0 +1,17 @@
...
+
+[Click here](https://example.com/javascript/about).
+
+To learn more, go [here!](https://example.com/javascript/about).
+
+To learn more, go to this [link!](https://example.com/javascript/about).
+
+<!-- markdownlint-configure-file {
+  "descriptive-link-text": {
+    "banned_link_texts_override": ["go here"]
```
DavidAnson, line 15: I love that you've added this test. Please have it verify two banned texts for added safety.

test/descriptive-link-text.md
```diff
@@ -0,0 +1,23 @@
+# Descriptive link text
```
DavidAnson, line 1: Please add another test file to confirm that setting the banned list to `[]` allows the default banned items - in other words, makes the rule allow everything.

test/descriptive-link-text-override-config.md
DavidAnson: I may do a final nitpick pass on your branch to make some of these silly changes myself instead of annoying you. :)

test/descriptive-link-text.md
```diff
@@ -0,0 +1,23 @@
...
+
+[Learn about Javascript](https://example.com/javascript/about)
```
DavidAnson, line 3: Please add some variety to the destinations. It should not matter, but one can be just "file.txt" and another can be "/dir/file.txt" or whatever. 10 instances of the same URL is not as interesting.

test/descriptive-link-text.md
DavidAnson: Yours is a reference link, but it looks more like a footnote. Let's use something more typical as shown in the examples I link.

test/descriptive-link-text.md
```diff
@@ -0,0 +1,23 @@
...
+
+[Read about Javascript](https://example.com/javascript/about)
+
+Learn about [JavaScript](https://example.com/javascript/about).
```
DavidAnson, line 7: Let's please have an example of every default banned text in this file. Again, it shouldn't matter, but it will alert if someone ever accidentally deletes or corrupts one.

test/long-lines-long-reference-definitions.md
```diff
@@ -1,5 +1,6 @@
 # Long Lines Long Reference Definitions
 
+<!-- markdownlint-disable-file descriptive-link-text -->
```
DavidAnson, line 3: Move to bottom of file, please.

DavidAnson (changes requested):
Looking good, thanks for making all those edits!! After going through these new comments, please update `schema/build-config-schema.mjs` which will automatically update some of the other files like the schema, etc., once you run `npm run ci`. I see test failures, but assume it's mostly stuff like this. You may get failures in the TestRepos workflow - use `npm run update-snapshots` and `npm run clone-test-repos` then `npm run update-snapshots-test-repos` to automatically update all the snapshot files. We can have a look at what impact this new rule has on the test repos based on that.

lib/md059.mjs
```diff
@@ -0,0 +1,64 @@
+// @ts-check
+
+import { addErrorContext } from "../helpers/helpers.cjs";
+import { filterByTypesCached } from "./cache.mjs";
+
+const defaultBannedNames = [
+  "click here",
```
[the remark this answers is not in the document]
khiga8: I don't have numbers behind this list. Some of these like "Click here" and "learn more" are commonly brought up as "bad" examples across various accessibility-focused resources so it was top of mind for me to include (examples: [Section 508 article on Descriptive links](https://www.section508.gov/blog/accessibility-bytes/descriptive-links-and-hypertext/), [W3: Don't say click here](https://www.w3.org/QA/Tips/noClickHere)). I suppose others of these are "vibes", or text I've seen out in the wild which made sense to include.

> I'm curious if an actual analysis would produce other interesting examples

That would be interesting!

For funsies, here are recurrences of each of these with GitHub code search across public repos:

1. [here - 2.4 million matching files](https://github.com/search?type=code&q=path%3A.md+%5Bhere%5D)
2. [click here - 197k matching files](https://github.com/search?type=code&q=path%3A.md+%5Bclick+here%5D0)
3. [learn more - 97.3k matching files](https://github.com/search?type=code&q=path%3A.md+%5Blearn+more%5D)
4. [read more - 89.6k matching files](https://github.com/search?type=code&q=path%3A.md+%5Bread+more%5D).
5. [more - 33.5k matching files](https://github.com/search?type=code&q=path%3A.md+%5Bmore%5D)

lib/md059.mjs
```diff
@@ -0,0 +1,60 @@
...
+      for (const labelText of labelTexts) {
+        const { text } = label;
+        if (bannedNames.has(normalizeText(text))) {
+          const range = labelText.startLine === labelText.endLine ?
+            [ labelText.startColumn, text.length ] :
+            undefined;
+          addErrorContext(
+            onError,
+            labelText.startLine,
+            text,
```
khiga8, line 51:
```
  -       errorContext: `[click␍␊
  +       errorContext: `[click␊
          here]`,
```

I can't figure out why [Windows behaves differently](https://github.com/DavidAnson/markdownlint/actions/runs/12840914345/job/35810130734?pr=1459) and returns a different `text` for the link test case with a line break. Curious if you might have any ideas @DavidAnson.

khiga8:
Hey @DavidAnson! I believe I've addressed all your comments now. I have one remaining issue I'd appreciate your eyes on - https://github.com/DavidAnson/markdownlint/pull/1459#discussion_r1920954182.

lib/md059.mjs
DavidAnson: This looks like a \r\n vs. \n line ending difference. If you're getting this error, it's reporting an error that spans multiple lines. I left a comment about preventing that by checking startLine and endLine of the span. (See my comment on the main chat before spending time on this.)

DavidAnson:
Thank you! I replied to that open issue above. (FYI, I may not get back to this for a couple of days.)

At this stage in the process, I'll make another/final pass over the entire change. It's likely I'll have more nitpicks. Are you okay with me applying them to your PR here? What I typically do is squash the author's comments into one (attributed to then) with a following commit by me titled something like "final tweaks for previous commit". Unless you enjoy my nitpicking, this is probably the most efficient option. :)

One thing I'll do is review every new violation in test-repos. Skimming that now, it looks like the vast majority are "here" and "link" which is expected (plus a couple of "more"s). However, there are about 20 like this in mdn which deserve better understanding. If you beat me to that, please reply here with your thoughts:
```
test-repos/mdn-content/files/en-us/web/html/attributes/rel/preload/index.md: 105: MD059/descriptive-link-text Link text should be descriptive [Context: "[\`<link>\`]"]
```

khiga8:
Hey @DavidAnson! Sorry for the delay here!

> (See my comment on the main chat before spending time on this.)
> At this stage in the process, I'll make another/final pass over the entire change. It's likely I'll have more nitpicks. Are you okay with me applying them to your PR here? 

Yep, I'm totally fine with you applying changes here! I haven't addressed the failing test yet - let me know if you'd like me to follow-up there, or with anything else.

> If you beat me to that, please reply here with your thoughts:

It looks like the linter is flagging link texts with `<link>`, which links to an [this article about the link HTML flag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link). In this scenario, `<link>` _actually_ describes the content of the linked page, and isn't used in a generic way, so we definitely don't want to flag this.

Right now, we're stripping all non-alphanumeric characters in order with the intent to flag texts like `link!` and `click here.`, but I'm thinking we'd want to be more conservative in this method, and only strip specific punctuations like `,!.:`.

DavidAnson:
In the mdn scenario, it's ``[`<link>`](/en-US/docs/Web/HTML/Element/link)`` which means there's a code span inside the link text. I think that scenario should be exempt from this rule, so will look at filtering such things out at a token level vs. via character replacement. This PR is next on my list, so I expect to spend some time on it this week. I'll take care of the above as part of my final pass on this. :)

DavidAnson (approved):
Accepting PR into a dedicated branch for final review.

PR #1564, opened 2025-04-14
Omit micromark parsing if there is no related rules

3y3:
In cases where all default markdownlint rules are disabled
and only custom rules based on markdownit tokens are used.

For example:
The Diplodoc platform uses 9 custom rules,
all of which are implemented using markdownit.
All default markdownlint rules are disabled
(we use markdownlint only as a rule execution engine).

In this configuration, using micromark for parsing is not only unnecessary
but also significantly slows down the linting process.

### DIFF
--- a/lib/markdownlint.mjs
+++ b/lib/markdownlint.mjs
@@ -490,12 +490,14 @@ function lintContent(
   const needMarkdownItTokens = enabledRuleList.some(
     (rule) => (rule.parser === "markdownit") || (rule.parser === undefined)
   );
+  const needMicromarkTokens = enabledRuleList.some(
+    (rule) => (rule.parser === "micromark")
+  );
   const customRulesPresent = (ruleList.length !== rules.length);
   // Parse content into parser tokens
-  const micromarkTokens = micromarkParse(
-    content,
-    { "freezeTokens": customRulesPresent }
-  );
+  const micromarkTokens = needMicromarkTokens ?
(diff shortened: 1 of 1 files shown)


DavidAnson:
Great idea, I had no idea anyone was doing this! Please see CONTRIBUTING.md and update the PR, base branch should be "next" for example.

Also, I wonder how to test this - for the corresponding markdown-it change I was able to infer the parser was not called via a custom plugin. I'd like something similar here.

https://github.com/DavidAnson/markdownlint/commit/ea8d596a9b6f97ccc27cd2e6bd8d992c7f150fd2

DavidAnson:
I think an approach to testing that would work for this change is to add another test much like the one in the commit I link above, but this time implement the markdown-it custom rule and have that import cache.mjs and verify it's empty. That should only be true if none of the micromark rules ran and called cache helper functions. This test should also verify the cache is not empty when micromark rules do run as a self-check.

krud491 (commented):
finaly

DavidAnson:
Ignore krud491's review, they are not a maintainer.

krud491:
Finish

XProudfootX (commented):
Great

DavidAnson:
Ignore XProudfootX's review, they are not a maintainer.

DavidAnson:
Between adding tests and fixing the base branch and not having push permission, I wasn't able to complete this PR as-is. The proposed commit is in `next` branch as a parent of the above-linked commit. Thanks for the suggestion!

Issue #1613, opened 2025-05-23
MD01/heading-increment does not take front matter titles into account as level 1 heading


afeblot:
With markdownlint 0.38 (markdownlint-cli 0.45.0, ) all rules enabled with the default config:

This document generates the expected error:

```md
# level 1

### level 3
```

**mdtest.md:3 MD001/heading-increment Heading levels should only increment by one level at a time [Expected: h2; Actual: h3]**

However, this one where the level 1 heading is replaced by the front-matter title does not:

```md
---
title: "level 1"
---

### level 3
```

Given **MD025/single-title** and **MD041/first-line-heading** do consider this front matter as a level 1 heading, I would expect MD01 to behave consistently and do so as well.

If it does not, I have no clean way the check that pages with a front matter title start with a level 2 heading.

DavidAnson:
MD001 is not appropriate for ensuring the first heading is an H2 whether or not YAML is present. For example, it does not trigger here: https://dlaa.me/markdownlint/#%25m%23%23%20Heading%0A

This test file should show how to achieve what you want using MD041: https://github.com/DavidAnson/markdownlint/blob/224987d727ebada36752074d2466a061c9e03ff0/test/md041-yaml-title-and-first-line-heading.md?plain=1

afeblot:
@DavidAnson , thanks for your quick response.

But it does not help: I confirm your example works;

With this config,

```yaml
first-line-heading:
  level: 2
```

this doc

```md
---
path: "/post"
date: "2012-06-21T10:14:00.000+02:00"
title: "First level heading"
---

## level 2
```

does not trigger a warning. Which is what I expect.

However,

```md
---
path: "/post"
date: "2012-06-21T10:14:00.000+02:00"
title: "First level heading"
---

### level 3
```

does not trigger a warning either. Whereas I would expect a warning here.

What I would like is that:
- the title is considered as the first heading in the page, and to be a level 1 heading (this is the default behavior with MD0025)
- Obviously, be allowed to have multiple level 2 headings.
- A warning if the next heading after the title is not a level 2 (don't we agree this should be reported by default by mardownlint?).

I'm able to achieve this with the below config:

```yaml
default: true

first-line-heading:
  level: 2
  front_matter_title: dontFindTitle
  allow_preamble: true

single-title:
  level: 1
  front_matter_title: dontFindTitle

no-duplicate-heading:
  siblings_only: true
```

This config does trigger a warning on

```md
---
title: "TEST-A"
[...]

PR #1617, opened 2025-05-27
MD001: Support front matter title as the first heading of the page

afeblot:
Consistently with rules 25 and 41, this brings support for front matter title as the first heading of the page, which will by default be considered as a level 1 heading, unless a level is explicitly configured.

Cf. https://github.com/DavidAnson/markdownlint/issues/1613

@DavidAnson, 
- This change does not break any of the 611 tests and provides the behavior I think is approppriate for MD001 goal with a front matter title.
- It will definitely report things which were not previously, i.e. we can consider it a breaking change, but as far as I saw, that is not an issue with Mardownlint, which does sometimes add new rules as well, and enabled by default.
- **Please let me know what you think about the rule change. If you're ok to move on, I'll add tests, schema and so on.**


DavidAnson:
Ah, yes. :) Someone who was bothered by that but wanted to keep this rule enabled could presumably customize it to provide an invalid front matter title name like we both did above for the other rules, right? With that single change, I think the behavior across their project will be the same as it is today.

afeblot:
@DavidAnson, here's my first iteration. I consider it complete, feel free to review.
Thanks

doc-build/md001.md
```diff
@@ -26,6 +26,15 @@ level at a time:
 ### Another Heading 3
 ```
 
+If [YAML](https://en.wikipedia.org/wiki/YAML) front matter is present and
+contains a `title` property (commonly used with blog posts), this rule treats
+that as a top level heading and will report a violation if the actual first
+heading is not a level 2 heading.
+To use a different property name in the front matter,
```
DavidAnson, line 33: Is this paragraph copied from a different rule? The spacing is weird here - should this line be part of the previous?

test/snapshots/markdownlint-test-scenarios.mjs.md
```diff
@@ -16178,6 +16178,87 @@ Generated by [AVA](https://avajs.dev).
       `,
     }
 
+## heading_increment-alt_title_level3_ko.md
```
DavidAnson, line 16181: As with your other PR, please revert all snapshot changes other than the one for the scenarios tests.

test/heading_increment-title_level3_ko.md
```diff
@@ -0,0 +1,5 @@
+---
+title: heading_increment-title_level3_ko
```
DavidAnson, line 2: "ko" => "ok", I think? :)

lib/md001.mjs
```diff
@@ -11,7 +11,12 @@ export default {
   "tags": [ "headings" ],
   "parser": "micromark",
   "function": function MD001(params, onError) {
-    let prevLevel = Number.MAX_SAFE_INTEGER;
+    let prevLevel = frontMatterHasTitle(
```
DavidAnson, line 14: Nitpick, but could you please assign the result of frontMatterHasTitle to its own const boolean? Just to clean up the ternary so it can be single-line and more obvious.

DavidAnson (changes requested):
Minor stuff here, thank you.

doc-build/md001.md
afeblot: Yes, copied from rule 25 doc. I updated the line endings. If this still does not comply with what you'd expect, can you please suggest the exact formatting you'd like?

test/heading_increment-title_level3_ko.md
afeblot: Err, no, I named it "ko" on purpose, meaning that this document is wrong and raises the MD001 error, as opposed to heading_increment-title_level2_ok which is ok because it does not raise it. 
If you don't agree with the idea of having "ko" in these names, the "ok" is meaningless as well in heading_increment-title_level2_ok, and I can just remove them all.
Please let me know what you prefer.

doc-build/md001.md
DavidAnson: In MD025, there's no break between the two sentences: https://github.com/DavidAnson/markdownlint/blob/224987d727ebada36752074d2466a061c9e03ff0/doc-build/md025.md?plain=1#L29

That's the convention.

test/heading_increment-title_level3_ko.md
DavidAnson: Instead of "ko" as a suffix, how about "good" and "bad" as used by some other test files?

afeblot: ok, done.

afeblot:
All requested changes done, and rebased. Hopefully ready for merge.
Now, the usual question: Do you have any idea when you'd intend to release a version with these PRs?
Thanks

PR #1730, opened 2025-08-21
feat: implement autofix for md029

### LINKED ISSUE #949: [feature request] MD029/ol-prefix: no quick fix available for ordered lists where order got mixed up
There's no quick fix available for ordered lists, where the order got mixed up, e.g. after a line has been cut.
I think that could improve the usability of the already great extension even further! :wink: 
See below screenshot:

image

ferrarimarco:
Add support for auto-fixing MD029 violations.

Close #949


DavidAnson (changes requested):
Interesting! I hadn't realized the foundation for this was already in place. A very quick scan of the code changes on my phone mostly makes sense. It looks like there are some test failures to fix. Running "npm run ci" should reproduce those locally. There are also scripts to update the two kinds of snapshot files which looks like it will need to be done. Finally, I'm a little bit worried about using the column of the list item prefix instead of the list item value which is what we actually want to change. I think this could be fairly easily addressed by updating getOrderedListItemValue to return the column and the text value in an object or array instead of just the value like it does today. Thank you for working on this!

ferrarimarco:
Thanks for the feedback. I'll address that.

ferrarimarco:
@DavidAnson Can you please take another look? I think I addressed your feedback. Thanks!

DavidAnson:
Merged as <https://github.com/DavidAnson/markdownlint/commit/371a35f62f969bcb49ec2970c2a83b4354fc1c13>, thank you!

PR #1903, opened 2025-12-28
feat: add/remove leading/trailing piples for MD055 errors (fixes #1902).

### LINKED ISSUE #1902: MD055 add fixes
Current state as at v0.40.0: `MD055` provides detection, but no fixes.

Proposal: Add the following behaviour

- On `expectedLeadingPipe && !actualLeadingPipe`: insert `|` ahead of leading token in first cell of row
- On `!expectedLeadingPipe && actualLeadingPipe`: remove `|` and any leading whitespace from first cell of row
- On `expectedTrailingPipe && !actualTrailingPipe`: insert `|` after trailing token in last cell of row
- On `!expectedTrailingPipe && actualTrailingPipe`: remove `|` and any trailing whitespace from last cell of row

Note: this would not guarantee passing `MD060` (`table-column-style`), which would need a seperate fix.

non-zero-sum:
Adds fix to MD055 as per #1902 

- On expectedLeadingPipe && !actualLeadingPipe: insert | ahead of leading token in first cell of row
- On !expectedLeadingPipe && actualLeadingPipe: remove | and any leading whitespace from first cell of row
- On expectedTrailingPipe && !actualTrailingPipe: insert | after trailing token in last cell of row
- On !expectedTrailingPipe && actualTrailingPipe: remove | and any trailing whitespace from last cell of row


DavidAnson:
Thanks for the contribution! I can't see much from my phone right now, but have two general comments:

How did you address your comment in the issue about fixes here introducing MD060 violations for its compact/tight styles? I don't see where this PR accounts for that. Also, rules typically work independently, so this strategy may need more thinking.

I'm not sure it's necessary to add 2x more flavors of every table column style to the test files, but if that's valuable, I prefer to do test-only changes like that in a dedicated commit (PR in this case) so the diffs for the following code change are clear.

non-zero-sum:
> How did you address [...] introducing MD060 violations [...] this strategy may need more thinking.

You're right this PR doesn't address it. It just naively adds/removes the pipe. I'll add my thoughts on this to the Issue.

>  Prefer to do test-only changes like that in a dedicated commit

Given the discussion needed on this one I will leave it for now and seperate out any test updates once the direction is clearer. 

Do you want to close this PR and I'll raise a new one after the discussion in the Issue?

Issue #1397, opened 2024-10-31
`.markdownlint.jsonc` config `json-schema.org` schema: allow trailing commas


jimmy-zhening-luo:
# Problem Statement

1. Edit a `.markdownlint.jsonc` file in `VSCode`.
2. Manually specify `$schema` as good practice:
  ```jsonc
  {
    "$schema": "https://raw.githubusercontent.com/DavidAnson/markdownlint/v0.36.1/schema/markdownlint-config-schema.json",
  }
  ```
3. Verify that trailing commas are allowed, because `markdownlint` happily accepts a `.jsonc` config with trailing commas.

## Expected

VSCode does not show any warning about trailing commas.

## Actual

VSCode shows warnings about trailing commas.

![image](https://github.com/user-attachments/assets/c5acd249-27f2-4625-a95a-30defd67364f)

![image](https://github.com/user-attachments/assets/0b5e01cf-c8f5-4c8c-9e2d-2516bf01da8a)

### Mitigation

1. In `.markdownlint.jsonc`, comment out `$schema` field:
  ```jsonc
  {
    // "$schema": "https://raw.githubusercontent.com/DavidAnson/markdownlint/v0.36.1/schema/markdownlint-config-schema.json",
  }
  ```
2. By default, VSCode complains about trailing commas in any `jsonc` file __unless told otherwise by a `$schema`__. To allow commas unles told otherwise by a `$schema`, add this to your VSCode user or workspace `settings.json`:
  ```jsonc
  {
    "json.schemas": [{
      "fileMatch": ["**/*.jsonc"] /* array of patterns to match, where here I prefer a block-list approach when it comes to .jsonc */,
      "schema": {
        "allowTrailingCommas": true,
      },
    }],
  }
  ```

Voila!

![image](https://github.com/user-attachments/assets/6fdfc629-ea83-4e7c-8daf-231c7f5d93c4)

And, miraculously, __VSCode still validates schema even without `$schema`__, either because of DavidAnson's extension is smartly doing some file association, or because VSCode Markdown Language Support is auto-magic (I haven't checked yet 😅, but I'm assuming the former because anyone fastidious enough to ship a `json-schema.org` schema would probably do that).

![image](https://github.com/user-attachments/assets/7a7fb9ee-f29a-4cdb-bdd5-8a52ae4df1db)

# Proposed Change

Create a `json-schema.org` schema for `.markdownlint.jsonc` with `{ "allowTrailingCommas": true }`.

[...]

PR #1939, opened 2026-02-06
feat: JSONC schema

Zamiell:
The current schemas do not allow trailing commas, which is a problem because in JSONC files, it is optimal to have trailing commas for all the same reasons that you would want trailing commas in your JavaScript/TypeScript files. Furthermore, tools like Prettier will automatically add trailing commas when auto-formatting JSONC files. Thus, I have added a new schema for people to use for JSONC files.

### DIFF
--- a/schema/ValidatingConfiguration.md
+++ b/schema/ValidatingConfiguration.md
@@ -19,8 +19,13 @@ npx ajv-cli validate -s ./markdownlint/schema/markdownlint-config-schema.json -d
 By default, any rule name is valid because of custom rules. To allow only
 built-in rules, use the
 [`markdownlint-config-schema-strict.json`][markdownlint-config-schema-strict]
-JSON Schema instead.
+schema instead.
+
+To allow trailing commas in a JSONC file, use the
+[`markdownlint-config-schema-jsonc.json`][markdownlint-config-schema-jsonc]
+schema instead. (This is based on the strict schema.)
 
 [json-schema]: https://json-schema.org
 [markdownlint-config-schema]: markdownlint-config-schema.json
 [markdownlint-config-schema-strict]: markdownlint-config-schema-strict.json
+[markdownlint-config-schema-jsonc]: markdownlint-config-schema-jsonc.json
--- a/schema/build-config-schema.mjs
+++ b/schema/build-config-schema.mjs
@@ -12,6 +12,8 @@ const schemaName = "markdownlint-config-schema.json";
 const schemaUri = `https://raw.githubusercontent.com/DavidAnson/markdownlint/v${version}/schema/${schemaName}`;
 const schemaStrictName = "markdownlint-config-schema-strict.json";
 const schemaStrictUri = `https://raw.githubusercontent.com/DavidAnson/markdownlint/v${version}/schema/${schemaStrictName}`;
+const schemaJsoncName = "markdownlint-config-schema-jsonc.json";
+const schemaJsoncUri = `https://raw.githubusercontent.com/DavidAnson/markdownlint/v${version}/schema/${schemaJsoncName}`;
 
 // Schema scaffolding
 const schema = {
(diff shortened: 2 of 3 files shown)


DavidAnson:
I already do not like the duplication that the "additionalProperties" property creates, but doubling the number of schemas again (as this change would do) seems too far.

But perhaps we can defer that discussion because it does not seem to me (searching only from my phone) that the "allowTrailingCommas" property you propose to introduce is a valid JSON schema property. I do not see it described in the core or validation documents here: https://json-schema.org/specification

If I'm overlooking something, please link to the relevant specification. Otherwise, the enforcement of that behavior seems to be coming from some other tool you are using.

DavidAnson:
It seems I've looked into this twice before:

https://github.com/DavidAnson/markdownlint/issues/1397

https://github.com/DavidAnson/markdownlint-cli2/pull/270

My position has not changed since then.

Zamiell:
Ok. Sorry for not finding that issue myself before doing this PR!

Would you be open to changing this to a docs PR then? I propose that we should put in the docs something like:

> Note that if you use our schemas with a JSONC configuration file, [Visual Studio Code](https://code.visualstudio.com/) will throw warnings on trailing commas. [Trailing commas](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas) are a special feature of JSONC that JSON does not have. It is popular to have trailing commas in JSONC files, especially if you are using tools like [Prettier](https://prettier.io/). It is possible to get rid of the Visual Studio Code warning by using a custom schema property called [`allowTrailingCommas`](https://github.com/microsoft/vscode/pull/203387#issuecomment-1909686444). However, since it is nonstandard, [we do not support it](https://github.com/DavidAnson/markdownlint-cli2/pull/270) in the official markdownlint repository. Instead, you can use [a 3rd party schema that "extends" the official one](https://raw.githubusercontent.com/complete-ts/complete/refs/heads/main/packages/complete-cli/schemas/markdownlint-config-schema-jsonc.json).

DavidAnson:
I prefer not to document behavior/workarounds for external tools unless it's a very common scenario. In this case, VS Code documentation itself discourages trailing commas in JSONC:

> When in the JSON with Comments mode, you can use single line (//) as well as block comments (/* */) as used in JavaScript. The mode also accepts trailing commas, but they are discouraged and the editor will display a warning.

https://code.visualstudio.com/docs/languages/json#_json-with-comments

Zamiell:
> I prefer not to document behavior/workarounds for external tools unless it's a very common scenario. 

Trailing commas is a very common scenario though. This is demonstrated through Prettier being the most popular formatting tool in the ecosystem and Prettier automatically adding trailing commas to JSONC files. And VS Code is the most popular editor/IDE in the ecosystem.

> In this case, VS Code documentation itself discourages trailing commas in JSONC:

It's true that the VS Code documentation discourages it but the ecosystem/community disagrees and has already demonstrated as such.

I know of two popular lint tools that allow for JSONC configuration files: CSpell and Knip. Both allow for trailing commas:
- [CSpell has "allowTrailingCommas": true](https://raw.githubusercontent.com/streetsidesoftware/cspell/main/cspell.schema.json)
- [Knip has  "allowTrailingCommas": true](https://unpkg.com/knip@5.83.1/schema-jsonc.json)

But perhaps there are other popular lint tools that I am not aware of that are counter-examples. Do you know of any? I'd be curious to know if any exist. (Besides Markdownlint, obviously.)

DavidAnson:
To be clear, this library does not prevent you from using trailing commas in configuration files. You are complaining about the way your editor is enforcing a valid JSON schema which this library provides for convenience.

Zamiell:
Sorry, I didn't mean to come across as complaining. My intent was to convince you that this falls in the realm of "a very common scenario", and should thus be called out explicitly in the markdownlint docs. If you agree, I can update the PR accordingly.

DavidAnson:
Both of the schemas you link above as examples that allow trailing commas do so because of PRs that YOU sent:

- https://github.com/streetsidesoftware/cspell/pull/4930
- https://github.com/webpro-nl/knip/pull/320

In the second case, the project owner resisted the idea before eventually giving in. Therefore, I do not think these are compelling examples.

Researching this situation further on my own, I found there is an open issue against prettier to STOP adding trailing commas to JSONC files:

https://github.com/prettier/prettier/issues/15956

And there are a LOT of people commenting there asking for a change, raising similar objections, and linking to problems prettier has caused for their projects.

As such, I am not interested in wading into an issue that ultimately has nothing to do with this project. JSON schemas are meant to define the structure of data, NOT its representation, and so it is inappropriate to introduce a representation-specific property like allowTrailingCommas.

I believe this project’s schema is correct as-is and do not wish to try to explain this mess in the documentation. Like many others in the prettier issue, I think prettier should change its current behavior because it is clearly problematic.

PR #1973, opened 2026-02-21
fix(md012): do not flag blank lines adjacent to headings (fixes #990).

### LINKED ISSUE #990: Allow MD012 (no-multiple-blanks) to work with MD022 (blanks-around-headings)
This was previously proposed in the context of #305, but was beyond the scope of that issue. A new issue was recommended at the time, but I never got around to create one. Until now, that is :smile:

**The problem:** currently setting [MD022](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md022)/blanks-around-headings to any value other than the default (1) can conflict with [MD012](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md012)/no-multiple-blanks, whose default value is 1.

In #305 the proposed workaround was to increase MD012/no-multiple-blanks's maximum to the number of lines specified for a heading; however, that makes it impossible to prevent the occurrence of multiple blank lines in other contexts.

One possible solution would be to have the values of MD022 override MD012's (similar to how in CSS more specific selectors override more generic ones); but perhaps that CSS analogy already hints at what would IMHO be a cleaner solution: since multiple blank lines around headings are a _special case_ of multiple blanks, it seems reasonable to add parameters to MD012 to allow specifying custom values in that specific context. Something like this, for example:

```yaml
no-multiple-blanks:
  maximum: 1
  headings_above: 2
  headings_below: 1
```

For prior art about rules that support parameters to apply differently in different contexts, see:
- [MD013/line-length](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md013), which supports the `code_blocks`, `headings` and `tables` parameters;
- [MD030/list-marker-space](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md030), which supports the `ol_multi`, `ol_single`, `ul_multi` and `ul_single` parameters
- [MD033/no-inline-html](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md033), which supports the `allowed_elements` parameter
- and various others.

I don't think we need additional parameters for other elements, but I'll mention the possibility nevertheless in case it may be considered a more consistent approach overall: this proposal of extending MD012 could subsume not just MD022, but also [MD028](https://github.com/DavidAnson/markdownlint/blob/main/doc/md028.md)/no-blanks-blockquote, [MD031](https://github.com/DavidAnson/markdownlint/blob/main/doc/md031.md)/blanks-around-fences, and [MD032](https://github.com/DavidAnson/markdownlint/blob/main/doc/md032.md)/blanks-around-lists.

/cc @wedi

rvben:
Fixes #990.

When `MD022` is configured with `lines_above` or `lines_below` greater than 1,
the required blank lines are also flagged by `MD012`, making the two rules
impossible to satisfy simultaneously.

This change pre-computes the blank lines immediately adjacent to each heading
and excludes them from `MD012`'s check. Those blank lines are governed by
`MD022`, so `MD012` defers to it. The adjacent-blank walk uses `isBlankLine()`
(the same helper `MD022` uses) so both rules count heading-adjacent blank lines
consistently.

A new test fixture covers the case, and the rule documentation notes that
heading-adjacent blank lines are governed by `MD022`.

Note: this PR was developed with AI assistance. I'm disclosing this in line
with the contributing guidelines and leaving it to your discretion whether
to accept it.


DavidAnson:
> Note: this PR was developed with AI assistance. I'm disclosing this in line with the contributing guidelines and leaving it to your discretion whether to accept it.

Thank you for calling this out! I don’t think this contribution includes code that is subject to a contradictory license, but the problem of using an LLM is that’s how they work - by design - and I can’t tell just by looking at the diff. I do not want to assume the risk of misusing/abusing someone else’s work, so I’ve banned the use of LLMs in `CONTRIBUTING.md` and do not accept their output.

PR #2028, opened 2026-04-07
feat(md033): add fix_strategy parameter

caugner:
Allows auto-fixing disallowed inline HTML elements.

We have two use cases on MDN:

1. When using machine translation (from English to German), escaped/wrapped HTML elements in prose occasionally gets translated into an unescaped/unwrapped HTML element, causing markdownlint to fail. Auto-fixing allows to resolve these without manual intervention.
2. When contributors forget to wrap HTML elements in prose, they have to manually fix these. Auto-fixing allows us to auto-suggest the most obvious fix (in our case: wrapping in a code fence) via our reviewdog integration.

### DIFF
--- a/doc-build/md033.md
+++ b/doc-build/md033.md
@@ -16,6 +16,11 @@ specific set of HTML elements within Markdown tables, set the
 `table_allowed_elements` parameter to a list of HTML element names. This can be
 used to permit the use of `<br>`-style line breaks only within Markdown tables.
 
+To sanitize disallowed elements automatically, set the `fix_strategy` parameter:
+
+- `"escape"`: replaces `<my-example>` with `\<my-example>`
+- `"wrap"`: replaces `<my-example>` with `` `<my-example>` ``
+
 Rationale: Raw HTML is allowed in Markdown, but this rule is included for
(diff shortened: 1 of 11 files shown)


DavidAnson:
Thanks! Before we go any further, did you use AI and/or an LLM for any part of this commit?

caugner:
I did use Claude Code to explore the repository, and draft the initial implementation, which I reworked before pushing the initial commit in this PR. I realize now that the paragraph in the contributing doc disallows the use of such tools broadly, so my contribution should be rejected.

For what it's worth, ad0c247c1e09b39bbb6c6a6deacf9c63d94d2533 does not contain any generated code.

DavidAnson:
@caugner, thank you for the update! Please note that you should be able to implement the behavior you want as a custom rule. Those behave just like normal rules but because the implementation lives outside this repository, you can do whatever you want in them and build them however you like. Some internal rules share code that's not directly available to custom rules, but I checked MD033 and the two helpers it uses are both pretty trivial, so I think it would not be hard to implement the necessary scaffolding in a custom rule to support the fix behavior you want.

Here's the documentation:

https://github.com/DavidAnson/markdownlint/blob/main/doc/CustomRules.md

PR #2124, opened 2026-07-02
MD029: fix corrupted output for zero-padded list markers

chatman-media:
The autofix works out how many characters to delete by re-stringifying the parsed number (`Number(text).toString().length`) instead of measuring the actual marker text. For a marker like `08.` that gives 1 instead of 2, so the fix only strips the leading `0` and leaves the rest behind — `08.` turns into `18.` instead of `1.`. Only shows up with leading zeros, which the MD029 doc explicitly calls out as supported for uniform indentation.

Fixed by keeping the source text length around instead of recomputing it from the number. Added a fixture with a zero-padded list that reproduces the corruption and confirmed the fixed output re-lints clean.

### DIFF
--- a/lib/md029.mjs
+++ b/lib/md029.mjs
@@ -15,12 +15,13 @@ const listStyles = Object.keys(listStyleExamples);
  * Gets the column and text of an ordered list item prefix token.
  *
  * @param {import("markdownlint").MicromarkToken} listItemPrefix List item prefix token.
- * @returns {{column: number, value: number}} List item value column and text.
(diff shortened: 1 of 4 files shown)


DavidAnson:
Thank you! Please review `CONTRIBUTING.md`, rebase this off of `next` branch, and say if you used an LLM to produce this PR. I'll review after that - I have some very minor feedback.

DavidAnson:
No reply, so updated myself with a more comprehensive fix for upcoming patch-level release. No PR code was used to avoid LLM/licensing concerns. Thanks for reporting the issue.

PR #2159, opened 2026-07-14
Add MD013 line-length fix check

### LINKED ISSUE #535: MD013: Automatically fix line length
When rule [MD013/line-length](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md013) is enabled and markdownlint is fixing rule violations, it would be helpful if it could reflow a multiline paragraph. This would greatly improve productivity when editing paragraphs broken at a specific column.

Gijsreyn:
Closes: #535.

This PR adds `--fix` support to MD013/line-length. Long paragraph lines are wrapped at whitespace, with wrapped lines prefixed to preserve block quote (> ) and list item (indentation) context.

Wrapping is conservative: it only applies to paragraph text (never headings, code, tables, or reference definitions) and never breaks inside code spans, inline HTML, autolinks, link destinations/titles, or reference labels — nor at any position that would change parsing (e.g., before text that would become a list item, heading, block quote, table row, or fence, or after a backslash). Lines with no safe break point still report a violation, just without a fix.

> [!NOTE]
> I know from the issue that this has been parked and had a pending redesign, but I thought, why not give it a try. I'll expect scrutiny 😄.

### DIFF
--- a/doc-build/md013.md
+++ b/doc-build/md013.md
@@ -35,5 +35,11 @@ of a paragraph) with only a link/image (possibly using (strong) emphasis) are
 always exempted from this rule (even in `strict` mode) because there is often no
 way to split such lines without breaking the URL.
 
+Long lines of paragraph text can be fixed automatically by wrapping them at
+whitespace. Wrapped lines are prefixed to preserve block quote and list item
+context. Wrapping does not occur at positions that would change the meaning of
+the content, such as within a code span or before text that would become a list
+item, so some long lines can not be fixed automatically.
+
 Rationale: Extremely long lines can be difficult to work with in some editors.
 More information: <https://cirosantilli.com/markdown-style-guide#line-wrapping>.
--- a/doc/Rules.md
+++ b/doc/Rules.md
@@ -497,6 +497,8 @@ Parameters:
 - `strict`: Strict length checking (`boolean`, default `false`)
 - `tables`: Include tables (`boolean`, default `true`)
 
+Fixable: Some violations can be fixed by tooling
+
 This rule is triggered when there are lines that are longer than the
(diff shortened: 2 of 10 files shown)


DavidAnson:
I appreciate your interest in solving challenges!

However, I have issues with the approach. It appears you did not read `CONTRIBUTING.md` and you also have not addressed the concerns I raised about this strategy in the discussion for issue #535.

For an example of why I claim this process is not practical, consider a typical violation introduced by an edit:

```text
Text text text text text
text text text text EDIT text
text text text text.
```

The proposed technique produces the following result which I claim is not what anyone wants:

```text
Text text text text text
text text text text EDIT
text
text text text text.
```

What I think people want is to reflow the entire paragraph - but I don’t believe that amount of change is appropriate for a lint tool.

You may disagree; if you’d like to create a custom rule, I’m sure others would appreciate it.

Gijsreyn:
@DavidAnson - thanks for taking the time to review this and for the feedback. And yes, apologies, you're right on that I didn't read the `CONTRIBUTING.md` before opening this. I see now that the expected process is to discuss the approach on the issue first (and to prototype new behavior as a custom rule), rather than showing up with code. That's on me.

Your example is a fair criticism of the greedy single-line wrap: pushing the overflow onto a short continuation line technically satisfies the rule but produces output nobody would write by hand.

Full paragraph reflow is what people actually want, and I agree that coordinating multi-line edits is beyond what the current single-line fix API (and a lint tool generally) should be doing — which is exactly the concern you'd already raised in #535.

I'll close this PR and may explore packaging the reflow behavior as a custom rule instead, where the more aggressive rewriting is opt-in. Thanks again for the response.

PR #2197, opened 2026-07-30
Accept one space per side in an empty cell for MD060 style "compact"

### LINKED ISSUE #2196: MD060: style "compact" requires one space per side for content cells but one space total for empty cells
## Summary

`MD060` style `compact` applies a different padding rule to empty cells than to cells with content. A cell with content needs one space on each side, while an empty cell is only allowed one space in total. `doc/md060.md` describes the style as "avoids extra padding with a single space around cell content", which reads as one space per side, so the empty-cell behaviour looks unintended.

## Reproduction

markdownlint 0.41.1, config `{"default": false, "MD060": {"style": "compact"}}`.

Cells with content, one space per side is required:

| cell | violations |
| :--- | :--- |
| `\|a\|` | 2, missing space to the left and right |
| `\| a \|` | 0 |
| `\|  a  \|` | 2, extra space to the left and right |

Empty cells, only one space in total is accepted:

| cell | violations |
| :--- | :--- |
| `\| \|` | 0 |
| `\|  \|` | 2, extra space to the left and right |
| `\|   \|` | 2 |

So `| a |` passes but `|  |` does not, although both have exactly one space between the pipe and the (empty) content on each side.

Full document that reproduces it:

```markdown
# T

|  |  |
| :--- | :--- |
| a | b |
```

```
line 3: MD060/table-column-style [Table pipe has extra space to the right for style "compact"]
line 3: MD060/table-column-style [Table pipe has extra space to the left for style "compact"]
(twice, once per empty cell)
```

`--fix` rewrites line 3 to `| | |`.

## Cause

In `lib/md060.mjs` the check walks `tableCellDivider`, `tableContent` and `whitespace` tokens and, for each divider, compares the adjacent `whitespace` token against a length of 1:

```js
if (previous.type === "whitespace") {
  if (previous.text.length !== 1) { /* extra space to the left */ }
```

For a cell with content there are two separate whitespace tokens, one on each side of the `tableContent` token, and each is checked once against 1.

For an empty cell there is no `tableContent` token, so micromark emits a single `whitespace` token spanning the whole gap. That one token is then checked twice, once as the right padding of the preceding divider and once as the left padding of the following divider, each time against a length of 1. A gap of two spaces, which is one space per side, therefore fails both checks, and a gap of three spaces produces the same two errors.

## Why it matters

Table formatters commonly implement padding as "one space, content, one space" applied uniformly to every cell, which yields two spaces for an empty cell. Output of that shape is internally consistent but MD060 reports it under every style, including `any`, and the fix rewrites it. Where such a formatter also owns the files, the two tools revert each other on every save.

In my case that is roughly 59,600 reported violations across 1,818 of 2,142 files, all from empty cells alone, in a Markdown documentation repository whose tables are formatted by an editor plugin. The only self-consistent alternative is to switch the formatter to no padding at all and use style `tight`, which costs a lot of readability for documents that are read as plain text and in diffs.

## Suggestion

Accept a two-space gap in an empty cell under `compact`, on the basis that it is one space per side and matches the documented wording. Continuing to accept the current one-space form would keep existing files valid.

If you would rather not change `compact`, an option on the rule or an additional style would also solve it.

I am happy to open a pull request for whichever shape you prefer, including doc and test updates. I did not send one straight away because this changes rule semantics and that seemed like your call to make first.

Spenhouet:
Fixes #2196.

## Problem

Style `compact` applies a different padding rule to empty cells than to cells with content. A cell with content needs one space on each side, an empty cell is only allowed one space in total:

| cell | before | after |
| :--- | :--- | :--- |
| `\| a \|` | 0 | 0 |
| `\|a\|` | 2, missing space | 2, missing space |
| `\|  a  \|` | 2, extra space | 2, extra space |
| `\| \|` | 0 | 0 |
| `\|  \|` | 2, extra space | 0 |
| `\|   \|` | 2, extra space | 1, extra space |

`doc/md060.md` describes the style as "avoids extra padding with a single space around cell content", which reads as one space per side, so `|  |` failing looked unintended.

## Cause

A cell with content has two `whitespace` tokens, one on each side of the `tableContent` token, and each is checked once against a length of 1.

An empty cell has no `tableContent` token, so micromark emits a single `whitespace` token spanning the whole gap. That one token was then checked twice, once as the right padding of the preceding divider and once as the left padding of the following divider, each time against a length of 1. A two-character gap, which is one space per side, therefore failed both checks, and a three-character gap produced the same two errors.

## Change

Precompute which `whitespace` tokens are bounded by `tableCellDivider` tokens on both sides. Such a token is the padding of an empty cell, so for style `compact` it may be up to two characters and it is reported once rather than once per side.

A single space stays valid, so existing files and existing `--fix` output are unaffected. A gap of three or more is still reported, and now fixes to two rather than one.

Cells with content are untouched, and style `tight` is untouched.

## Verification

`npm test` passes, 677 tests, `lib/md060.mjs` at 100% coverage. `npm run lint` is clean, and `npm run serial-config-docs` regenerated `doc/md060.md` and `doc/Rules.md`, both committed.

Notably the snapshot diff is confined to the fixture section I added, so nothing in the existing corpus changed behaviour. Added coverage to `test/table-column-style-compact.md` for a two-space empty cell, a one-space empty cell, an empty cell in a mixed row, and a three-space gap.

## Why it matters

Table formatters commonly implement padding uniformly as "one space, content, one space", which yields two spaces for an empty cell. Output of that shape is internally consistent but was reported under every style, including `any`, and the fix rewrote it. Where such a formatter also owns the files, the two tools revert each other on every save. In my repository that was roughly 59,600 violations across 1,818 of 2,142 files from empty cells alone.

Happy to adjust the shape if you would rather express this as an option or a separate style than as a change to `compact`.

### DIFF
--- a/doc-build/md060.md
+++ b/doc-build/md060.md
@@ -30,6 +30,18 @@ Style `compact` avoids extra padding with a single space around cell content:
 | N | No |
 ```
 
+An empty cell has no content, so a single space around that content is a gap of
+two characters. Both that and a single space are accepted for style `compact`,
+because table formatters differ in which of the two they emit:
+
+```markdown
+| Character | Meaning |
+| --- | --- |
+| Y | Yes |
+|  |  |
+| | |
+```
+
 Style `tight` uses no padding at all for cell content:
 
 ```markdown
--- a/doc/Rules.md
+++ b/doc/Rules.md
@@ -2773,6 +2773,18 @@ Style `compact` avoids extra padding with a single space around cell content:
 | N | No |
 ```
 
+An empty cell has no content, so a single space around that content is a gap of
+two characters. Both that and a single space are accepted for style `compact`,
+because table formatters differ in which of the two they emit:
+
+```markdown
+| Character | Meaning |
+| --- | --- |
+| Y | Yes |
+|  |  |
+| | |
+```
+
 Style `tight` uses no padding at all for cell content:
 
 ```markdown
--- a/doc/md060.md
+++ b/doc/md060.md
@@ -44,6 +44,18 @@ Style `compact` avoids extra padding with a single space around cell content:
 | N | No |
 ```
 
+An empty cell has no content, so a single space around that content is a gap of
+two characters. Both that and a single space are accepted for style `compact`,
+because table formatters differ in which of the two they emit:
+
+```markdown
+| Character | Meaning |
+| --- | --- |
+| Y | Yes |
+|  |  |
+| | |
+```
+
 Style `tight` uses no padding at all for cell content:
 
 ```markdown
--- a/lib/md060.mjs
+++ b/lib/md060.mjs
@@ -119,13 +119,28 @@ export default {
         }
         for (const row of rows) {
           const tokensOfInterest = filterByTypes(row.children, [ "tableCellDivider", "tableContent", "whitespace" ]);
+          // An empty cell has no tableContent token, so its padding is a single
+          // whitespace token between two dividers. Style "compact" allows one
+          // space on each side of cell content, so such a gap may be one or two
+          // characters, and it is reported once rather than once per side.
+          /** @type {Set<number>} */
+          const emptyCellGaps = new Set();
+          for (let i = 1; i < (tokensOfInterest.length - 1); i++) {
+            if (
+              (tokensOfInterest[i].type === "whitespace") &&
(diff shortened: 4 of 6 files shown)


DavidAnson:
See #2196

Issue #11, opened 2016-02-09
Rule MD004 incorrectly flagging on an Ordered List


gryte:
I'm getting flagged on the following. Am I missing something obvious?

```
1. Fork the repository on Github
+ Create a named feature branch (like `add_component_x`)
+ Write your change
+ Write tests for your change (if applicable)
+ Run the tests, ensuring they all pass
+ Submit a Pull Request using Github
```

DavidAnson:
Are you on the latest version of markdownlint?

I pasted your example into the top-left box of http://dlaa.me/markdownlint/ and do not see any MD004 violations.

Could you please try that?

gryte:
Interesting...yeah, I see what you mean if I use the website you referenced (nice by they way).

I confirmed that I'm on markdownlint version 0.3.0 which seems to be the latest in the marketplace.

I feel like I'm not using the correct md syntax to make an ordered list, but when I try the following, I get MD029 in both code and the website.

```
1. Fork the repository on Github
2. Create a named feature branch (like `add_component_x`)
3. Write your change
4. Write tests for your change (if applicable)
5. Run the tests, ensuring they all pass
6. Submit a Pull Request using Github

```

Do you have a good reference for appropriate syntax handy?

Thanks for the help!

DavidAnson:
You saying that version 0.3.0 is the latest in the marketplace makes me think you're referring to the markdownlint VS Code extension https://github.com/DavidAnson/vscode-markdownlint (which uses this markdownlint engine). However, I've tried your original example with that and still do not see the error you're reporting.

The reason your second example triggers MD029 is because the default configuration for MD029 is to prefer the style 1./1./1. over 1./2./3. because the latter is easier to get wrong. This is discussed in the relevant documentation page: https://github.com/DavidAnson/markdownlint/blob/master/doc/Rules.md#md029---ordered-list-item-prefix You can change the default in the usual manner - here's an example: https://github.com/DavidAnson/markdownlint/blob/master/test/ordered_list_item_prefix_ordered.json For details on making that change for vscode-markdownlint, please see: https://github.com/DavidAnson/vscode-markdownlint#configure

Regarding documentation about syntax, I like the following:
* http://daringfireball.net/projects/markdown/
* http://commonmark.org/

gryte:
Sorry I failed to mention I was in vs code. Thanks for the clarification!

DavidAnson:
[...]

PR #2208, opened 2026-08-05
Improve MD060 wide-character coverage (arrows, rare CJK, fullwidth forms)

seeones:
## Background

MD060 (`table-column-style`) computes each table column's visual width with the `string-width` package. In practice, arrows such as `↑ ↓ ← →` (U+2190–U+21FF) and some wide characters outside the common CJK ranges (rare ideographs like `𠀀`, `〇`, vertical forms, fullwidth signs, etc.) are not always measured as double-width, so "aligned" tables containing them are still reported/mis-fixed.

## Changes

- `lib/md060.mjs`: replace the `string-width` import with a local `visualWidth()` helper — grapheme-cluster count via `Intl.Segmenter` (Node ≥16, well within the required Node ≥22) plus an explicit UAX \#11 *Wide/Fullwidth* character class. The class additionally covers arrows (U+2190–U+21FF), which are "Ambiguous" in UAX \#11 but render double-width in CJK contexts/terminals.
- `test/table-column-style-wide-characters.md`: add `Arrows` and `Rare CJK` sections; `test/markdownlint-test-scenarios.mjs.snapshot` updated accordingly.

To keep this PR minimal, `package.json` is intentionally untouched — `string-width` is no longer imported anywhere and its dependency entry can be dropped in a follow-up at the maintainer's discretion.

## Testing

- `node --test test/markdownlint-test-scenarios.mjs` (408 tests) and the remaining snapshot suites (119 tests) pass.
- `eslint --max-warnings 0 lib/md060.mjs` passes.

## Related

The same width-calculation approach is used by the table formatter in `vscode-markdown` (Markdown All in One), which had the identical coverage gap. A corresponding fix has already been submitted there and is open for review:

- PR: https://github.com/yzhang-gh/vscode-markdown/pull/1581


DavidAnson:
If the `string-width` package is handling these characters incorrectly, please open an issue with that project. That will improve all projects that use it vs. requiring separate changes everywhere.

DavidAnson:
See https://github.com/DavidAnson/markdownlint/issues/2219 for an example where `string-width` gave the correct behavior vs. another tool that did the wrong thing because it had an incomplete implementation.

